PR #33, opened 2016-06-26
Fix install_svn and related functions

dmurdoch:
This fixes the bug in install_svn() that stops it from working with R-forge.  It no longer edits the URL, adding "trunk" or "branches" onto it.  

Since this makes it incompatible with the original, I also simplified things so they make more sense from an SVN point of view:  I dropped the "branch" argument, used "export" instead of "checkout", and chose the revision at time of export, rather than exporting the head and updating afterwards.

### DIFF
--- a/R/install-svn.R
+++ b/R/install-svn.R
@@ -9,35 +9,32 @@
 #'
 #' @inheritParams install_git
 #' @param subdir A sub-directory withing a svn repository that may contain the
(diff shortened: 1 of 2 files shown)


gaborcsardi:
Thanks! THis is great! Don't worry about the appveyor failure, that seems to be a separate issue.

PR #36, opened 2016-09-09
Fix tests on OSX and Linux (Travis)

gaborcsardi:


gaborcsardi:
Appveyor still fails because of a different issue.

PR #46, opened 2016-11-27
Change default download method if needed

### LINKED ISSUE #45: download package on windows behind proxy does not work (easily)
Hi, 

Trying to solve this issue with `devtools::install_github` (hadley/devtools#1403), I found that `remotes` package is probably the next thing for some functions in  `devtools`. So I tried it. 

However, I encountered an issue as I am on windows and behind a proxy. Trying to install a github package throws an error
```r
remotes::install_github("hadley/lubridate")
#> Error in utils::download.file(url, path, method = download_method(), quiet = quiet,  :  
#> cannot download all files
 ```
and even crash R session if proxy is not configured correctly.

This issue is clearly proxy-related.

below is what I have done to make it work. See at the end for Suggestions

### About proxy configuration

Proxy configuration is not the easy thing but I managed to deal with it most of the time. [Configuring R to Use an HTTP or HTTPS Proxy](https://support.rstudio.com/hc/en-us/articles/200488488-Configuring-R-to-Use-an-HTTP-or-HTTPS-Proxy) by RStudio support team helps. 

However, the easiest way on windows is to rely on `wininet` methods which is the default for `download.file`. see in windows help:

> If method = "auto" is chosen (the default), on Windows the "wininet" method is used apart from for ftps:// URLs where "libcurl" is tried. The "wininet" method uses the WinINet functions (part of the OS).

`wininet` deals with proxy using the 'Internet Option' of the system. And it is used even if `libcurl` is available (`capabilities("libcurl")` is `TRUE`)

In `remotes`, by default `remotes:::download` calls `remotes:::download_method` in which the default is `libcurl` if it is available. Then, if not, it is `wininet`

At first, default behavior threw an error. With some efforts I manage to make it work by configuring proxy differently. 
- The way describe in [Configuring R to Use an HTTP or HTTPS Proxy](https://support.rstudio.com/hc/en-us/articles/200488488-Configuring-R-to-Use-an-HTTP-or-HTTPS-Proxy) does not work anymore as it is correct for `internal` only. 
- For `libcurl`, the form to use is `http[s]://[user:password@]machine[:port]`. Downloading part of `remotes` package is now working correctly.
- For`httr` package, I uses `set_config` and `use_proxy` to make it work. (Needed for `devtools` then as it is working with `httr`) but I think the `http_proxy` environnement variable is also working

### Suggestion 

At the end, I found how to deal with the issue. However, I think it could be useful to clarify all this about proxy configuration
- By making `wininet` the default on windows before testing `capabilities("libcurl")` in `remotes:::download_method`
- If `libcurl` stays the default, by adding some informations in a vignette or in help file to help those behind a proxy. And maybe try to prevent errors or crashes.

What do you think of all this ? I willing to help if you interested as I planned to write something about all this configuration.

### PR DESCRIPTION
**Pull request by cderv**
This PR follows a discussion to resolve #45 . See issue post.

Based on some research about downloads method in R, I decided to modify `download_method` function in the following way

```
if R is older than 3.3, 
    if OS is windows
        set method to "wininet" 
    else on other OS if libcurl is avalaible, 
        set method to "libcurl"

For R 3.3 or newer, set method to "auto" the default
```

- Comparing with the previous function, I just reversed the checking (Windows before libcurl) and I rely on default R for newer than 3.3

- I noticed that for R older than 3.2.0, `curl_download` is used and `dowload_method` is not called. So I did not take care of this case. 
- If libcurl  is not available on non windows O, we rely on default method 

- I have modified the test to be compliant with the new function, and I have added some infos in Readme about download methods to point users toward `download.file` function help file.

- I did not change `intall-github.R` file as it seems you generated it by brew. 
- I did not add a bullet point in NEWS.md as it seems you have not done it before yet.

We could have done differently
* Modify default only for https url like in `downloader` package
* Mixed `base_download` and `curl_download` and deal with all that in `download` function as it is done in `downloader` function

Hope it is not too bad and I did not forgot anything
Tell me if something is not clear in what I have done

### DIFF
--- a/R/download.R
+++ b/R/download.R
@@ -50,15 +50,18 @@ download_method <- function() {
   if (compareVersion(get_r_version(), "3.3") == -1) {
     
     if (os_type() == "windows") {
-      return("wininet")
+      "wininet"
+      
+    } else if (isTRUE(unname(capabilities("libcurl")))) {
+      "libcurl"
+      
+    } else {
+      "auto"
+    }
     
-      } else if (isTRUE(unname(capabilities("libcurl")))) {
-      return("libcurl")
-    
-      }
+  } else {
+    "auto"
   }
-  
-  "auto"
 }
 
 curl_download <- function(url, path, quiet) {
--- a/R/download.R
+++ b/R/download.R
@@ -45,16 +45,20 @@ base_download <- function(url, path, quiet) {
 }
 
 download_method <- function() {
-
-  if (isTRUE(unname(capabilities("libcurl")))) {
-    "libcurl"
-
-  } else if (os_type() == "windows") {
-    "wininet"
-
-  } else {
-    "auto"
+  
+  # R versions newer than 3.3.0 have correct default methods
+  if (compareVersion(get_r_version(), "3.3") == -1) {
+    
+    if (os_type() == "windows") {
+      return("wininet")
+    
+      } else if (isTRUE(unname(capabilities("libcurl")))) {
+      return("libcurl")
+    
+      }
   }
+  
+  "auto"
 }
 
 curl_download <- function(url, path, quiet) {
--- a/inst/README.md
+++ b/inst/README.md
@@ -127,6 +127,18 @@ installed temporarily.
   and archived packages as well.
 * All dependencies of a package in a local directory via
   `install_deps`.
+  
+### Download methods
+
+* For R older the 3.2, `curl` package is required as `remotes` fallbacks to
+`curl::curl_download` in that case
+* For R newer than 3.3, default `download.file` method is used. (`method = "auto"`)
+* For in between version, 
+    * `method = "wininet"` is used on windows OS
+    * `method = "libcurl"` is used on other OS, if available.
+
+See `help("download.file")` for informations on these methods and for setting
+proxies if needed.
 
 ### Notes
 
--- a/tests/testthat/test-download.R
+++ b/tests/testthat/test-download.R
@@ -2,21 +2,29 @@
 context("Download")
 
 test_that("download_method", {
-
+  
   with_mock(
-    `base::capabilities` = function(...) c(libcurl = TRUE),
-    expect_equal(download_method(), "libcurl")
+    `remotes::get_r_version` = function(...) "3.3.0",
+    expect_equal(download_method(), "auto")
   )
 
   with_mock(
-    `base::capabilities` = function(...) c(libcurl = FALSE),
+    `remotes::get_r_version` = function(...) "3.2.5",
     `remotes::os_type` = function() "windows",
     expect_equal(download_method(), "wininet")
   )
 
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: R/download.R [hunk 1]
```diff
@@ -45,16 +45,20 @@ base_download <- function(url, path, quiet) {
...
+    if (os_type() == "windows") {
+      return("wininet")
+    
+      } else if (isTRUE(unname(capabilities("libcurl")))) {
+      return("libcurl")
+    
+      }
   }
+  
+  "auto"
```
gaborcsardi: Can you please remove the `return()` calls, and put in an `else` for the `"auto"`?

gaborcsardi:
THanks! Left a small comment, otherwise it looks good.

cderv:
Done. Just a prevision : If "libcurl" is not available on non windows platform, method is set to "auto".

Now, not sure what I should do. If you have some advices...
Do I need to squash the three commits into one ? 
Will you do this when merging? 
Are we staying with three commits ? 

I look forward to collaborate more to learn more  on the way to go with PR and projects collaborations

gaborcsardi:
Thanks, and sorry for the delay! You don't need to squash, GH can squash-and-merge now. Thanks again!

cderv:
Hello, 

I just noticed that I did not modified `install-github.r` file that you use for installing the dev version from github without any other package (ie `devtools::install_github`). I know your working on a new service "install-me" but should we update `download_method` function in `install-github.r` until it is available ? 
I am not sure how this file is generated. 

So, just for you to know, PR is merged and it is working now on windows but 
```r
source("https://raw.githubusercontent.com/MangoTheCat/remotes/master/install-github.R")$value("mangothecat/remotes")
```
is not working without proxy configuration on windows as it has not been updated with the new version of `download_method`.

gaborcsardi:
Good point. I'll update the script, and the `install-github.me` service as well. 

I should probably decide if I want 
```r
source("https://raw.githubusercontent.com/MangoTheCat/remotes/master/install-github.R")$value("mangothecat/remotes")
```
or 
```
source("https://install-github.me/mangothecat/remotes")
```

gaborcsardi:
I updated the script and the script and the service as well. Can you please try that both commands in the previous comment work? Thanks!

cderv:
I tried and it is working now - both method.

I would have done it but I was not sure if your `install-github.r` script is generated automatically or if I could update it manually in the different repo.

gaborcsardi:
No worries, it is updated like this:
```
Rscript -e 'brew::brew("install-github.Rin", "install-github.R")'
```
or the equivalent code from R, with the working directory set to the project root.

And you cannot update the service, because of permissions.

cderv:
OK. good to know how it is working, I was wondering !  Thanks !

cderv:
Thank you for the change notice @gaborcsardi ! I appreciate it.

PR #68, opened 2017-06-26
Fix install_local(".")

### LINKED ISSUE #67: install_local(".")
Fails for a simple package:

``` r
system("git clone https://github.com/krlmlr/tic.packagedocs.git")
withr::with_dir("tic.packagedocs", remotes::install_local("."))
#> Error: Don't know how to decompress files with extension yml
```

Succeeds with directory name as `path` argument:

``` r
system("git clone https://github.com/krlmlr/tic.packagedocs.git")
withr::with_temp_libpaths(remotes::install_local("tic.packagedocs"))
#> Installing package into '/tmp/Rtmpk1WkuT/temp_libpath7d12c71a3d7'
#> (as 'lib' is unspecified)
```

I can work around with `getwd()`, but it would be great if this just worked.

krlmlr:
by calling `normalizePath()` when constructing the remote. Closes #67.

Do we need a test?

### DIFF
--- a/R/install-local.R
+++ b/R/install-local.R
@@ -23,7 +23,7 @@ install_local <- function(path, subdir = NULL, ...) {
 
 local_remote <- function(path, subdir = NULL, branch = NULL, args = character(0)) {
   remote("local",
-    path = path,
+    path = normalizePath(path),
     subdir = subdir
   )
 }


gaborcsardi:
i think it is good.

gaborcsardi:
Thanks!

PR #74, opened 2017-07-12
Remove duplicate function definitions

krlmlr:
I haven't checked if the dupes are actually identical.

I wonder if there's a check for this kind of error. Shouldn't be too difficult to implement.

### DIFF
--- a/install-github.R
+++ b/install-github.R
@@ -3,7 +3,7 @@ function(...) {
 
   ## This is the code of the package, put in here by brew
 
-  
+
 bioc_version <- function() {
   bver <- get(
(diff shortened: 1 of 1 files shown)


gaborcsardi:
I don't understand this PR. Did you just auto-generate `install-github.R`?

krlmlr:
`safe_install_packages()` and `missing_devel_warning()` are dupes already defined in `install.R` .

gaborcsardi:
The file you modified is not part of the R package. It is only used for self-installing `remotes` and the `install-github.me` service.

krlmlr:
Missed that part, sorry.

PR #103, opened 2017-08-03
Initial support for git submodules

rtobar:
This initial support is based on the work done by @jonkeane for devtools, plus minor additions from my side. It includes submodules support for git2r, xgit and github remotes. It is not fully recursive yet, and the code is still not all cleanly separated, but it works as a first approach.

This pull request addresses #82.

### DIFF
--- a/R/install-git.R
+++ b/R/install-git.R
@@ -8,6 +8,7 @@
 #' @param url Location of package. The url should point to a public or
 #'   private repository.
 #' @param branch Name of branch or tag to use, if not master.
+#' @param submodules if \code{TRUE} download submodules before installing.
 #' @param subdir A sub-directory within a git repository that may
 #'   contain the package we are interested in installing.
 #' @param git Whether to use the \code{git2r} package, or an external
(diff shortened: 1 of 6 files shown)


ronammar:
Thanks for this useful feature @rtobar and @jonkeane! Any idea when this code will be merged into master?

rtobar:
No idea from my side. The pull request has been hanging out for a bit, but nothing much has happened since. Hopefully your comment will bring the attention back to this pull request.

ronammar:
Bump?

gaborcsardi:
Well, you said that:

> not fully recursive yet, and the code is still not all cleanly separated,

so I was wondering if you wanted to finish it.

ronammar:
@rtobar - thoughts? I've been using your fork and it works recursively for me, so what functionality is absent?

rtobar:
Hi @ronammar, @gaborcsardi 

When I say that it's not fully recursive I mean that submodules are not checked for submodules (e.g., repository A references repository B, which in turn references repository C), which IIRC hasn't been implemented yet. I didn't even try that scenario, so if it works it's pure chance. When you say that it works recursively for you, do you mean exactly that, or only that one level of submodule references works fine?

By the code not being cleanly separated I mean that the specifics about "recursiveness" are not generic enough at the moment since {{install-remotes.R}} has code that is specific to {{git}}-related remotes, instead of generalizing this concept and coding it into each of the supported remotes. As a first iteration it works though, which is why I though this request was maybe already good enough to be present as a first iteration.

If @gaborcsardi thinks this needs a bit more of work before being accepted I could try to find some time to address the comments, but can't promise any speedy delivery.

ronammar:
@rtobar Admittedly, I've only tested with a single submodule level, so it behaved the same as a `git clone` recursive call for the same repo.

ronammar:
Tested and it works for me. Thanks!

PR #108, opened 2017-09-13
Accept common URLs as repo spec in parse_github_repo_spec(); closes #90

### LINKED ISSUE #90: install_github() could accept full URL to GitHub repository
<a href="https://github.com/kevinushey"></img></a> **Issue by [kevinushey](https://github.com/kevinushey)**
_Thursday Sep 01, 2016 at 15:45 GMT_
_Originally opened as https://github.com/hadley/devtools/issues/1324_

----

Sometimes, when I am exceedingly lazy, I just try to copy + paste the path to a GitHub repository into `install_github`. It would be 🔥🔥🔥 if `install_github` were able to magically parse that.

jennybc:
PLEASE DO NOT MERGE THIS. As discussed, I'd like to make this PR live from my R-Ladies thing tomorrow night. But let's discuss if this solution is OK. Once true, I will close this and make the real PR then. I just did it in the simplest way, w/o disturbing the world's scariest regex just below 😃


R/install-github.R
```diff
@@ -235,12 +239,21 @@ github_resolve_ref.github_release <- function(x, params) {
...
 #' parse_github_repo_spec("jimhester/covr#47")
 #' parse_github_repo_spec("hadley/dplyr@*release")
 #' parse_github_repo_spec("mangothecat/remotes@550a3c7d3f9e1493a2ba")
+#' parse_github_repo_spec("https://github.com/jimhester/covr")
+#' parse_github_repo_spec("https://github.com/jeroen/curl.git")
+#' parse_github_repo_spec("git@github.com:metacran/crandb.git")
 
 parse_github_repo_spec <- function(repo) {
+
+  repo <- gsub("\\.git$", "", repo)
```
gaborcsardi: I would only remove the `.git` suffix with the prefixes, because some git urls might not handle the `.git`-less form, like GitHub.

gaborcsardi:
Thanks! Otherwise it looks great, I have wanted this for a long time!

gaborcsardi:
You can submit the exact same PR if you want, and I'll be online to request the change. :)

gaborcsardi:
and a bullet point in NEWS...

R/install-github.R
jennybc: OK I can do that. But just to check ... what are you actually worried about? That there might be a `.git` I should *not* strip? That there might be a GitHub repo actually named `foo.git` and `jenny/foo.git` should return username = `jenny` and repo = `foo.git`?

jennybc:
See what you think now.

I don't immediately see why travis r-devel is failing now but passed initially for this PR.

NEWS.md
```diff
@@ -1,6 +1,9 @@
 
 # 1.1.0
 
+* Accept browser, HTTPS, or SSH URL in GitHub repo specification,
```
gaborcsardi: What do you mean by "browser"?

gaborcsardi:
Actually, you are right, chopping off the `.git` is fine. I was just confused and thought that chopping off the `.git` might break other remote types (because it is used for bitbucket for example, quite confusingly), but you are right, it would not break anything.

So feel free to submit the original or the updated form.

NEWS.md
gaborcsardi: But how is that different from just a HTTPS url?

gaborcsardi: Or any URL within the repo is supposed to work? Like this one works as well: `https://github.com/r-lib/remotes/pull/108`?

jennybc:
> So feel free to submit the original or the updated form.

I think it does feel more contained the way it is now, so better.

I will close this and make a fresh one sometime between 7:30 and 8:10pm tonight. But I'm still going to point to this, because it's ended up being a good example of a PR process!

gaborcsardi:
Sorry, one more question. URLs like `https://github.com/r-lib/remotes/pull/108` don't work, right? Or they do?

jennybc:
No that would currently give the wrong result. The only "browser URL" that explicitly works is the main one `https://github.com/USERNAME/REPO`. Anything else would be a happy coincidence.

``` r
library(remotes)
parse_github_repo_spec("https://github.com/r-lib/remotes/pull/108")
#> $username
#> [1] "r-lib"
#> 
#> $repo
#> [1] "remotes"
#> 
#> $subdir
#> [1] "pull/108"
```

I thought about this. There are many forms (pull requests, branches, etc.) but didn't want to tackle all of them. I still think the current addition is worth but, yeah, don't want to overpromise re: browser URL support.

gaborcsardi:
How about just cutting off that part?

gaborcsardi:
At least for now?

jennybc:
Maybe I should just mention the HTTPS and SSH URLs in the help and let the other be an undocumented things that works.

We are typing at same time, but is that basically what you mean?

gaborcsardi:
I mean, I would just cut off the rest of the URL, i.e. `https://github.com/r-lib/remotes/pull/108#issuecomment-329440222` would be interpreted as `https://github.com/r-lib/remotes` 

Although this is not a good idea if we want to implement sg special for the various urls. 

OK, I think it would be better to give an error in this case. So basically, the URL has to end with the repo name.

You don't need to do this for today, but I think in the long term it would make sense. 
We'll add it to the new giant regex later :)
https://github.com/r-lib/depends/blob/20b52bb2c9d1680f02cb2d10dc75af1f9a946dc4/R/parse-remotes.R#L6

jennybc:
Got it.

gaborcsardi:
Great! Thanks much!

jennybc:
> OK, I think it would be better to give an error in this case. So basically, the URL has to end with the repo name.

I've added this, i.e. the only browser-style URLs that work are the HTTPS remote URL and the top-level URL. Otherwise, they error.

I also made it work for GitHub enterprise URLs, since I see that `install_github()` supports that.

How does this look?

gaborcsardi:
Great! Thanks much!

gaborcsardi:
Do you want to close this one?

jennybc:
Taken care of, as planned, in #109.

PR #110, opened 2017-09-19
Use re_match() in parse_github_repo_spec()

jennybc:
I plan to make a PR for handling the other forms of GitHub URL, such as for a commit, branch, release or pull. But it might be nice to have access to rematch2. And, in that case, maybe we could enjoy it here first?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -18,8 +18,9 @@ Description: Download and install R packages stored in 'GitHub',
 License: GPL (>= 2)
 URL: https://github.com/r-lib/remotes#readme
 BugReports: https://github.com/r-lib/remotes/issues
-Imports:
+Imports: 
     methods,
+    rematch2,
     utils
 Suggests:
     crancache,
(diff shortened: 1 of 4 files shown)


DESCRIPTION
```diff
@@ -18,8 +18,9 @@ Description: Download and install R packages stored in 'GitHub',
 License: GPL (>= 2)
 URL: https://github.com/r-lib/remotes#readme
 BugReports: https://github.com/r-lib/remotes/issues
-Imports:
+Imports:
```
jennybc: `usethis::use_package()` seems to add this space 😕

gaborcsardi:
I this is cool. Two things.

1. we are rewriting this functionality currently, in `pkgdepends`, and other `pkg*` packages. I am not yet sure what the fate of `remotes` will be, probably will just wrap the other packages.
2. in the meanwhile, `remotes` should be ideally dependency free. This is needed to be self-installing (see README) and also for the `install-github.me` service. 

So, I don't mind experimenting with feature now, but could you please avoid the `rematch2` requirement? If you only need `re_match`, then you can copy over the function from `rematch2`, it should be standalone I think. Thanks!

jennybc:
OK I copied `re_match()` in.

I understand the situation re: `pkg*` diaspora. I will still see this experiment through (the URL stuff). I think it will be nice to be able to install from any ref, PRs, etc. with URLs vs a handcrafted spec. This only matters for interactive human use, of course. Hopefully this little bit of UI can be moved to the correct place when the dust settles.

gaborcsardi:
Thanks!

Issue #13, opened 2016-01-08
Expose remotes:::parse_git_repo


richfitz:
Any chance this could be exposed?  I have a couple of use cases for it (building drat repositories mostly, but it's come up a few times).  Understand if you want to keep the package super simple though.

gaborcsardi:
Hmmm. In general I don't really want to export the little pieces like this. It is quite small, so I think you can just take the code, if you are fine with the license.

jennybc:
I have exactly the same request. How do I resolve the fact that the same code appears here (MIT) and in devtools (GPL) and originated in devtools?

gaborcsardi:
@jennybc My mistake, then `remotes` must be GPL, too.

gaborcsardi:
@jennybc Just changed the license, sorry for the confusion. I guess I just assumed that like most of Hadley's stuff, `devtools` was MIT.

On the other hand, `remotes` is not using any code from R-core AFAIK, and since the code was mostly written by Hadley and me, I guess he could licence that code with MIT and then `remotes` could be MIT.

@hadley ?

jennybc:
Now sort of sorry I asked but I suppose it had to get straightened out.

Back to the original topic, `remotes:::parse_git_repo` is little but very useful in many contexts. I wholeheartedly :+1: @richfitz's request to export it.

gaborcsardi:
@jennybc @richfitz can you please tell me very briefly how you would use this? 

I am thinking that maybe another package would be a better place for it, some github-related package.

richfitz:
I currently use a parser in [drat.builder](https://github.com/richfitz/drat.builder) for fetching a collection of gh packages into a drat (e.g. [here](https://github.com/richfitz/drat)).

I did a similar thing in [dockertest](https://github.com/traitecoevo/dockertest) though I'll probably re-implement that on top of drat.builder because it deals better with the dependency sort on installation. It doesn't use devtools-style strings yet but it should.

I am also using a parser in [context](https://github.com/dide-tools/context) for specifying sources of non-CRAN packages for later installation on a remote computer.  It works internally by building a file-based drat repo.

An aside, if you did make it public you'd be able to extend the syntax beyond github so that:

* `github:MangoTheCat/remotes` and `MangoTheCat/remotes` were equivalent
* `bitbucket:username/repo` would work with bitbucket.

Or `github://` or whatever.

gaborcsardi:
[...]

PR #112, opened 2017-09-20
Parse GitHub URLs

jennybc:
The promised PR that allows `install_github()` to accept a variety of GitHub URLs. It should be merged after #110, because this branch works off that base. Or, I guess, I could close that, because those commits are included here as well.

I took the liberty of doing a general tidy of GitHub remote parsing:

  * Split GitHub repo parsing into two exported functions: `parse_repo_spec()` (the thing we've always had) and `parse_github_url()`. They are documented together.
  * Moved the dropping of empty elements into the (unexported) `parse_git_repo()` to make the return value of the above functions more predictable.
  * Put all related tests in their own file because it's nice to be able to do `test(filter = "parse")` w/o doing the full `install_github()` tests. 

~~I know it still needs a NEWS bullet. I'll write one with correct issue/PR numbers if this is good to merge.~~


jennybc:
To make installation via URL work for, say, GHE remotes, the parsed list would have to gain a slot for something about the host 🤔.

NAMESPACE
```diff
@@ -35,7 +35,8 @@ export(install_url)
 export(install_version)
 export(local_package_deps)
 export(package_deps)
-export(parse_github_repo_spec)
+export(parse_github_url)
+export(parse_repo_spec)
```
gaborcsardi: It is an incompatible change to drop `parse_github_repo_spec`, I think we should keep that name (as well?).

jennybc: OK I will address this. Need to re-orient myself to what I was doing!

Just for context, I can't find any evidence of `parse_github_repo_spec()` usage on CRAN, outside of remotes itself.

GitHub-wise, I see one other use:

  * https://github.com/mrc-ide/provisionr/blob/d74f2cd3fc3b59e878439ac8d766b83570b0a681/R/remote.R#L37 by @richfitz

gaborcsardi: Thanks!

richfitz: As requested here: https://github.com/r-lib/remotes/issues/13 😀

jennybc: OK I went back to the old name, @gaborcsardi.

Here's what I think (?) I was thinking:

  * Start to de-emphasize the GitHub-ness, since these repo spec parsers are not too hard to generalize to BitBucket, GitLab, and the like. But it's easy to make this change now and cross that bridge later.

gaborcsardi:
Thanks! Can we just make `parse_github_repo_spec` an alias to the new function? Then we don't need two implementations. I understand that the new function is doing more, but that is fine.

R/install-github.R
```diff
@@ -221,79 +221,108 @@ github_resolve_ref.github_release <- function(x, params) {
...
+    params$repo <- gsub("\\.git$", "", params$repo)
+  }
+  if (params$release == "latest") {
+    params$release <- "*release"
+  }
+
+  params[grepl("^[^\\.]", names(params))]
 }
 
 parse_git_repo <- function(repo) {
```
jennybc: Should this be exported?

jennybc: OK let's leave it alone for now then.

gaborcsardi:
Great, thanks!

PR #116, opened 2017-11-09
Ignore .Rbuildignore files for install_local()

### LINKED ISSUE #107: Avoid copying .Rbuildignore files in install_local()
They may be huge, e.g., revdepcheck artifacts.

krlmlr:
Seems to work for me, but what kind of tests do we need here? Should this be used for other methods, such as installing from GitHub?

Fixes #107.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,7 +26,7 @@ Suggests:
     git2r,
     mockery,
     pingr,
-    testthat,
+    testthat (>= 2.0.0),
     withr
 Depends:
     R (>= 3.0.0)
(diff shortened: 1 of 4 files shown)


gaborcsardi:
I think it is worth to have tests for the functions in rbuildignore.R.

krlmlr:
Added tests for two low-level workers. No integration tests yet, do we need them?

krlmlr:
@gaborcsardi: Would you mind taking another look? This is another minor pain when doing revdepchecks.

gaborcsardi:
Can you pls import `setNames`?

gaborcsardi:
Thanks!

gaborcsardi:
I had to revert this, because it is causing problems: 

```
> remotes::install_local("s2_0.2-1.tar.gz", dependencies = TRUE, INSTALL_opts = "--build")
Error in normalizePath(target, winslash = "/", mustWork = TRUE) : 
  path[1]="/tmp/RtmpbFSWLk/filed25b209f8d//home/docker": No such file or directory
Calls: <Anonymous> ... remote_download -> remote_download.local_remote -> normalizePath
Execution halted
```

PR #123, opened 2018-01-30
connect parts of the markdown hyperlink

wibeasley:


gaborcsardi:
IS the newline ok in the link?

wibeasley:
I deleted that newline (between the square brackets and the parentheses) because I think that's the problem.  Is that your question?

## [master](https://github.com/r-lib/remotes) (with the newline): 
image


## [patch](https://github.com/wibeasley/remotes/blob/c2a4d28b06f8b2c264161fea6acee40574569349/inst/README.markdown) (after removing the newline):
image

gaborcsardi:
@wibeasley thanks, yes, sorry, what I am asking, the link will still point at the right place, because there is a newline in it.

wibeasley:
(Sorry, I may be misunderstanding the question.  I know you have a lot more experience developing packages on GitHub than I do.)

I believe that the gap/newline is *not* ok (assuming you want 'Writing R extensions' to be a hyperlink to the url).

I deleted the gap between the `]` and the `(` so that when the user clicks 'Writing R extensions', it will take them to the url.  As it is now (and in that screenshot above for 'Master'), clicking 'Writing R extensions' does nothing.  It's just text enclosed in square brackets.

gaborcsardi:
Thanks!

PR #125, opened 2018-02-08
Ignore .Rbuildignore for install_local()

krlmlr:
I forgot that `install_local()` is also supposed to work for `.tar.gz` files. Fixed now.

Updates (compared to merged and reverted state): https://github.com/r-lib/remotes/compare/a01d3c14324...krlmlr:f-%23107-rbuildignore-2

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,7 +26,7 @@ Suggests:
     git2r,
     mockery,
     pingr,
-    testthat,
+    testthat (>= 2.0.0),
     withr
 Depends:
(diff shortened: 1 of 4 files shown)


gaborcsardi:
CAn't we just use pkgbuild for this? We would make it optional.

krlmlr:
Do you mean *pkginstall*?

jimhester:
We now (optionally, enabled by default) build the packages before installing, so I don't believe this is needed any longer.

Issue #106, opened 2017-08-22
Credentials in install_git


robertdj:
In `devtools::install_git` I can supply credentials that are passed on to `git2r::clone`.
In `remotes::install_git`, the credentials are not passed on...

PR #127, opened 2018-02-15
pass credentials to install_git (#106)

dleutnant:
With this PR, `remotes::install_git` accepts credentials like `devtools::install_git`.

### DIFF
--- a/R/install-git.R
+++ b/R/install-git.R
@@ -10,6 +10,7 @@
 #' @param branch Name of branch or tag to use, if not master.
 #' @param subdir A sub-directory within a git repository that may
 #'   contain the package we are interested in installing.
+#' @param credentials A git2r credentials object passed through to clone.
 #' @param git Whether to use the \code{git2r} package, or an external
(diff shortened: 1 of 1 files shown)


gaborcsardi:
Thanks! Can you please add a warning (or error?) for the credentials not working with the command line git?

dleutnant:
I‘m not sure. Could you give me additional information?

gaborcsardi:
`install_git()` can use either git2r or the system git. By default it chooses one automatically. This PR only implements credentials for git2r, but not for the system git. So if you specify the credentials, and remotes chooses system git (or you choose system git explicitly), then the credentials will be ignores. It makes sense to give a warning, or maybe even error in this case, I think.

dleutnant:
OK, I get it. Thanks for for clarifying. I'll take care of it.

jimhester:
The internals of the install functions changes some since this PR was written, so I added this in https://github.com/r-lib/remotes/pull/146.

Thank you for the PR!

PR #135, opened 2018-07-31
Add SHA updating logic for github remotes

jimhester:
This updates remotes with the logic from the current devel devtools (only for the github remote).

Since pkgman and friends seem to be a ways off it might be worth trying to update remotes further and use that on travis / appveyor etc. The process was somewhat easier than I expected originally.


R/github.R
```diff
@@ -27,10 +28,24 @@ github_commit <- function(username, repo, ref = "master") {
...
+github_DESCRIPTION <- function(username, repo, subdir = NULL, ref = "master", host = "api.github.com", ...) {
+
+  url <- file.path(paste0("https://", host),
+                   "repos", username, repo, "contents", paste0(subdir, "DESCRIPTION"))
+  url <- paste0(url, "?=", URLencode(ref))
+
+  tmp <- tempfile()
+  download(tmp, url, auth_token = github_pat())
+
+  rawToChar(base64enc::base64decode(gsub("\\\\n", "", fromJSONFile(tmp)$content)))
```
jimhester: This is the current blocker for this, we need to decode the base64 encoded content, so I guess would have to write a R only version of `base64decode()`.

jimhester: Update: now have this

yihui:
Yeah!! 🎉 🎉 🎉

jimhester:
@gaborcsardi I realize this is probably too big now to review in detail; do you have any overall comments and would you be ok with me merging this and working on the open issues?

gaborcsardi:
Ill take a look later today.

R/install-remote.R
```diff
@@ -58,3 +81,126 @@ is.remote <- function(x) inherits(x, "remote")
...
+      sha = x$RemoteSha))
+}
+
+#' @export
+format.remotes <- function(x, ...) {
+  vapply(x, format, character(1))
+}
+
+#' @export
+`[.remotes` <- function(x,i,...) {
```
gaborcsardi: Why is this trick needed? Can you  please add  it  as  internal docs, or  a comment?

R/install-git.R
```diff
@@ -90,6 +90,54 @@ remote_metadata.git2r_remote <- function(x, bundle = NULL, source = NULL) {
...
 }
 
+#' @export
+remote_package_name.git2r_remote <- function(remote, ...) {
+
+  tmp <- tempfile()
+  on.exit(unlink(tmp))
+  description_path <- paste0(collapse = "/", c(remote$subdir, "DESCRIPTION"))
+
+  # Try using git archive --remote to retrieve the DESCRIPTION, if the protocol
```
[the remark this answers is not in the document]
jimhester: You need command line git to get the remote package name, but if you can't get it because command line git is not available the thing we then just install the remote unconditionally. So it will still work without command line git, just won't be as efficient as it could be.

I think it is still useful to have both types of remotes, although I agree it does increase the complexity considerably; the code is already written...

R/install-git.R
```diff
@@ -90,6 +90,54 @@ remote_metadata.git2r_remote <- function(x, bundle = NULL, source = NULL) {
...
+  # or server doesn't support that return NA
+  res <- try(silent = TRUE,
+    system_check(git_path(),
+      args = c("archive", "-o", tmp, "--remote", remote$url,
+        if (is.null(remote$ref)) "HEAD" else remote$ref,
```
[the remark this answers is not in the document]
jimhester: Probably `HEAD` as it is most generic

R/install-remote.R
jimhester: It is possible it is not needed, I will remove it and see if anything breaks.

jimhester:
Thank you for the review Gabor, I appreciated the comments!

PR #137, opened 2018-08-09
Support building packages before installation

### LINKED ISSUE #71: Should install_local() call R CMD build first?
Some packages seem to require it, https://github.com/r-lib/remotes/issues/67#issuecomment-311596532.

### LINKED ISSUE #80: install_github(local = FALSE) ?
<a href="https://github.com/gaborcsardi"></img></a> **Issue by [gaborcsardi](https://github.com/gaborcsardi)**
_Wednesday Mar 23, 2016 at 16:52 GMT_
_Originally opened as https://github.com/hadley/devtools/issues/1126_

----

This was discussed in #655, but I think there is a new benefit now. With the `Remotes` field we can now define dependencies in GH packages, which is great. But we cannot specify that they should be installed with `local = FALSE`. 

This is a problem for some of our new packages, that'll have a lot of JS code that I don't really want to commit to the repo. I would ideally just download this code at build time. But then I cannot use `Remotes` to depend on them.

So, again, how about making `local = TRUE` for GH packages?

jimhester:
This is not completely done, I still need to add tests I guess, but I wanted to get some feedback on the approach first.

Also as it stands I don't think it completely fixes #80, as right now the argument is not passed to remotes.

Fixes #71
Fixes #80


jimhester:
@gaborcsardi this is now ready, if you get a chance I would appreciate a review!

R/install.R
```diff
@@ -44,6 +56,45 @@ safe_install_packages <- function(...) {
...
+      manual <- TRUE
+    }
+    build_opts <- build_opts[!(has_no_vignettes | has_no_manual)]
+    pkgbuild::build(pkgdir, dest_path = dest_path, binary = FALSE,
+      vignettes = vignettes, manual = manual, args = build_opts, quiet = quiet)
+  } else {
+    # No pkgbuild, so use tools:::.build_packages, which is what is called by R
+    # CMD build. Calling it directly avoids needing to start a new R CMD build
+    # process (which varies across OSs) and deal with shell quoting (which also
+    # varies across OSs)
```
gaborcsardi: This is not actually what the code does. Maybe these comments are wrong?

jimhester: Sorry yes, forgot to remove these

jimhester:
Thanks for the review!

PR #138, opened 2018-08-10
Submodules

jimhester:
This still needs some tests, but it works for simple cases, e.g. `install_github("https://github.com/jonkeane/mocapGrip")`

### DIFF
--- a/R/submodule.R
+++ b/R/submodule.R
@@ -1,0 +1,111 @@
+parse_submodules <- function(file) {
+  if (grepl("\n", file)) {
+    x <- strsplit(file, "\n")[[1]]
+  } else {
+    x <- readLines(file)
+  }
+
+  # https://git-scm.com/docs/git-config#_syntax
+  # Subsection names are case sensitive and can contain any characters except
+  # newline and the null byte. Doublequote " and backslash can be included by
+  # escaping them as \" and \\
+  double_quoted_string_with_escapes <- '(?:\\\\.|[^"])*'
+
+  sections <- regexpr(
+    sprintf('^[[:space:]]*\\[submodule "(%s)"\\][[:space:]]*$', double_quoted_string_with_escapes),
+    x,
+    perl = TRUE)
+
+  # If no sections found return the empty list
+  if (all(sections == -1)) {
+    return(list())
+  }
+
+  # Otherwise extract section names
+  section_names <- get_captures(x, sections)
+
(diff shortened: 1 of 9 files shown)


gaborcsardi:
@jimhester Is this ready?

jimhester:
@gaborcsardi I was adding some tests, it is ready now if you could review it!

jimhester:
Oh I guess the only other thing this could maybe have is a git2r version, right now it relies on having a command line git only. I can add it if you think we need it.

gaborcsardi:
I think it is fine with system git.

R/submodule.R
```diff
@@ -0,0 +1,111 @@
...
+    x <- readLines(file)
+  }
+
+  # https://git-scm.com/docs/git-config#_syntax
+  # Subsection names are case sensitive and can contain any characters except
+  # newline and the null byte. Doublequote " and backslash can be included by
+  # escaping them as \" and \\
+  double_quoted_string_with_escapes <- '(?:\\\\.|[^"])*'
+
+  sections <- regexpr(
```
gaborcsardi: I wonder if we should just include a collated rematch2 package, that would simplify things.

tests/testthat/test-submodule.R
```diff
@@ -0,0 +1,113 @@
...
+  # url = https://foo/bar'
+
+  expect_equal(
+    parse_submodules(z),
+    list()
+  )
+})
+
+test_that("Can install a repo with a submodule", {
+  dir <- tempfile()
```
gaborcsardi: Need to skip if no system git.  (If we are skipping the other tests that rely on system git.)

gaborcsardi:
I think it is good now. 

I think if we want to be completely correct we would need to use the `GIT_DIR`  (etc.) env vars. But we'll probably never run into this. So maybe parsing the output of `git  submodule list` etc. is a better approach? 

I  don't think we need to change it now, just sg. to think about next time we write sg for git.

jimhester:
I think the issue with that is the zipball we get from GitHub is not actually a git repository, e.g. it doesn't have a `.git` directory. That is why I can't just run `git submodule init`, `git submodule update` to retrieve the submodules and have to parse them by hand.

gaborcsardi:
Got it. Yeah, these would only matter for local installs. All good then!

jimhester:
Thanks for the review!

PR #139, opened 2018-08-18
Use HTTPS for security

ankane:
🔒

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -377,7 +377,7 @@ fix_repositories <- function(repos) {
 
   # Override any existing default values with the cloud mirror
   # Reason: A "@CRAN@" value would open a GUI for choosing a mirror
-  repos[repos == "@CRAN@"] <- "http://cloud.r-project.org"
+  repos[repos == "@CRAN@"] <- download_url("cloud.r-project.org")
   repos
 }
 
--- a/R/download.R
+++ b/R/download.R
@@ -72,3 +72,40 @@ curl_download <- function(url, path, quiet) {
 
   curl::curl_download(url, path, quiet = quiet, mode = "wb")
 }
+
+true_download_method <- function(x) {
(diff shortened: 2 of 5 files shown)


gaborcsardi:
The problem with this is that some older R builds do not support HTTPS. Typically R 3.1.x, which we still want to support.

ankane:
What do you think of a conditional approach, where >= 3.2 uses HTTPS?

gaborcsardi:
Yes, but we would need to condition on actual HTTPS support. Or libcurl. This is with HTTPS support:
```
> capabilities("libcurl")
libcurl
   TRUE
```

Without can be FALSE, or even:
```
> capabilities("libcurl")
named logical(0)
```

gaborcsardi:
Although windows will support HTTPS  w/o libcurl, but AFAIR newer windows builds also link to libcurl, so it might be ok to ignore that.

ankane:
Cool, made it conditional on libcurl.

jimhester:
Unfortunately whether a given download method is secure is more complicated than just checking if R is compiled with libcurl support. Even in that case they could not be using libcurl and using a download method that is not insecure. Devtools has the following to determine if a download method is secure.

See the [devtools/R/download-method.R](https://github.com/r-lib/devtools/blob/26c507b128fdaa1911503348fedcf20d2dd30a1d/R/download-method.R).

gaborcsardi:
I think it is ok if we miss some. In practice  most new R builds will default to libcurl, and HTTPS will be probably fine for these.

gaborcsardi:
We could also copy over that logic from devtools.

ankane:
Updated PR based on your feedback @jimhester

gaborcsardi:
I think this is good. @jimhester ?

jimhester:
Thanks!

PR #142, opened 2018-08-20
Use faster retrieval of remote information for GitHub remotes

jimhester:
`remote_package_name.github_remote` and `remote_sha.github_remote` will
lookup package name locally if possible and use faster REST APIs if the curl
package is installed. This makes the remote lookup about 7-10x faster when the
package to be installed has not changed since the last install.

    pkgload::load_all("~/p/remotes")
    remote <- github_remote("r-lib/devtools")
    bnch <- bench::press(
      use_curl = c(FALSE, TRUE),
      use_local = c(FALSE, TRUE),
      bench::mark(
        get_shas = {
        package_name <- remote_package_name(remote, use_local = use_local, use_curl = use_curl)
        local_sha <- local_sha(package_name)
        remote_sha <- remote_sha(remote, local_sha, use_curl = use_curl)
      }, iterations = 3,
        filter_gc = FALSE
      )
    )
    #> Running with:
    #>   use_curl use_local
    #> 1 FALSE    FALSE
    #> 2 TRUE     FALSE
    #> 3 FALSE    TRUE
    #> 4 TRUE     TRUE

    bnch[c(2:10)]
    #> # A tibble: 4 x 9
    #>   use_curl use_local     min    mean  median   max `itr/sec` mem_alloc
    #>   <lgl>    <lgl>     <bch:t> <bch:t> <bch:t> <bch>     <dbl> <bch:byt>
    #> 1 FALSE    FALSE     563.8ms 606.8ms 586.1ms 670ms      1.65    4.65MB
    #> 2 TRUE     FALSE     160.9ms 273.5ms 165.6ms 494ms      3.66  206.08KB
    #> 3 FALSE    TRUE      269.7ms 273.1ms 269.9ms 280ms      3.66   19.37KB
    #> 4 TRUE     TRUE       74.1ms  86.5ms  84.7ms 101ms     11.6    19.12KB
    #> # ... with 1 more variable: n_gc <dbl>
    summary(bnch, relative = TRUE, filter_gc = FALSE)[c(2:10)]
    #> # A tibble: 4 x 9
    #>   use_curl use_local   min  mean median   max `itr/sec` mem_alloc  n_gc
    #>   <lgl>    <lgl>     <dbl> <dbl>  <dbl> <dbl>     <dbl>     <dbl> <dbl>
    #> 1 FALSE    FALSE      7.61  7.01   6.92  6.65      1       249.     Inf
    #> 2 TRUE     FALSE      2.17  3.16   1.96  4.90      2.22     10.8    NaN
    #> 3 FALSE    TRUE       3.64  3.16   3.19  2.78      2.22      1.01   NaN
    #> 4 TRUE     TRUE       1     1      1     1         7.01      1      NaN


jimhester:
@gaborcsardi My main question with this is the `use_local` argument. This assumes by default that if a package is installed locally with the same name as the remote repo name, then that is the remote repo's package name. I think this will be correct the vast majority of the time, but it could be a false positive if the remote repo is actually a different name. Therefore I think we need to have a user visible way to turn it off, but I think it will be rare enough I am not sure it is worth complicating the regular API to provide a regular option for it. What do you think about just letting it be set by an environment variable or global option?

gaborcsardi:
So, the thing is, in pkgdepends  (and co.)  if the package name is not the same as the repo name, you need to include the package name in the remote name.  It looks like this: https://github.com/r-lib/pkgdepends/blob/ed31a93b514c023f80ba617dedc2672f837d8271/tests/testthat/test-parse-remotes.R#L167-L170

We could maybe just include this in remotes as well? Maybe together with your heuristic, so there  won't be too much breakage?

jimhester:
Yeah that sounds good to me!

jimhester:
@gaborcsardi Any additional comments on this?

R/install-github.R
```diff
@@ -41,33 +39,27 @@
 #' install_github("hadley/private", auth_token = "abc")
 #'
 #' }
-install_github <- function(repo, username = NULL,
+install_github <- function(repo,
```
gaborcsardi: Removing  the  `username`  argument should be the NEWS, I think.

R/parse-git.R
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  if (!is.null(params$release)) {
+    params$ref <- github_release()
+    params$release <- NULL
+  }
+
+  params
+}
+
```
gaborcsardi: Is this all just the pkgdepends code?  Or new code?  Dont' mind, just curious.

jimhester: It is code from remotes, I just moved it out of `install-git.r` to a separate file to make testing it separately easier.

R/github.R
```diff
@@ -49,13 +62,34 @@ github_pat <- function(quiet = TRUE) {
...
-  base64_decode(gsub("\\\\n", "", fromJSONFile(tmp)$content))
+    curl::handle_setheaders(h, .list = headers)
+    res <- curl::curl_fetch_memory(url, handle = h)
+    if (res$status_code >= 300) {
+      stop("HTTP error ", res$status_code, ".", call. = FALSE)
+    }
+    rawToChar(res$content)
+  } else {
+    tmp <- tempfile()
+    on.exit(unlink(tmp))
```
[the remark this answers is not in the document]
jimhester: You only need `add = TRUE` if there is another `on.exit()` call in the same scope no? I guess `add = TRUE` unconditionally is safer, I can add it.

jimhester:
I think the worst that could happen if we assume the wrong name is the local sha would be from the wrong package, so the package would potentially be re-installed when it may not have had to be.

R/parse-git.R
gaborcsardi: OK, np.

gaborcsardi:
So this is all good now, I think, right?

jimhester:
Thank you for the review!

PR #143, opened 2018-08-27
Use gzip tarballs instead of zip archives when possible

jimhester:
These are generally smaller and should preserve execute permissions in
all cases, whereas some unzip implementations do not do so.

Fixes https://github.com/r-lib/devtools/issues/1799
Fixes https://github.com/r-lib/remotes/issues/96

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,11 @@
 
 # Development
 
+* `install_()` functions now download tarballs (.tar.gz) files rather than zip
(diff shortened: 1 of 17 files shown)


jimhester:
`git_extract_sha1_zip()` is not used anymore, I kept it around (with a test) just in case there was a backend that did not support tar.gz, but we could also just remove it entirely.

jimhester:
Thanks!

PR #145, opened 2018-08-28
Pass all authentication and build arguments to dependencies

### LINKED ISSUE #53: upgrade=FALSE is not passed through to dependent installations
Upgrading packages tends to lead to really broken sets of packages on windows due to file locking, and so I tend to set `upgrade = FALSE` quite a lot (I would *love* a `remotes.upgrade` global option that I could set to FALSE).

If a package installed by `remotes::install_github` lists a remote in its description then subsequent installations do not honour the `upgrade = FALSE` argument

### LINKED ISSUE #86: install_bitbucket fails when dependencies are private
<a href="https://github.com/moosilauke18"></img></a> **Issue by [moosilauke18](https://github.com/moosilauke18)**
_Wednesday Jun 29, 2016 at 19:38 GMT_
_Originally opened as https://github.com/hadley/devtools/issues/1245_

----

So, I have a repo in bitbucket which is private the depends on another private repo. As of the latest devtools version (1.12.0), it fails with a 401 unauthorized. I'm using `dependencies=TRUE` and sending auth_user and password in `install_bitbucket`. 

Also, the dependency I can install no problem first, so it's an issue when the dependency is private.

jimhester:
We now pass all authentication arguments and build arguments to remotes
and dependencies. This means you can supply the authentication
in in the top level `install_()` call and have that same authentication
be used in downstream Remotes.

This also should propagate upgrade to dependencies appropriately, so
setting `upgrade = FALSE` should disable upgrading for all dependencies,
even grandchildren.

Fixes #86
Fixes #53

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -282,19 +282,37 @@ install_packages <- function(packages, repos = getOption("repos"),
                              type = getOption("pkgType"), ...,
                              dependencies = FALSE, quiet = NULL) {
 
+  # We want to pass only args that exist in the downstream functions
+  args_to_keep <-
+    unique(
+      names(
+        c(
+          formals(install.packages),
+          formals(download.file)
+        )
+      )
+    )
+
+  args <- list(...)
+  args <- args[names(args) %in% args_to_keep]
+
   if (is.null(quiet))
     quiet <- !identical(type, "source")
 
   message("Installing ", length(packages), " packages: ",
     paste(packages, collapse = ", "))
 
-  safe_install_packages(
-    packages,
-    repos = repos,
-    type = type,
-    ...,
-    dependencies = dependencies,
-    quiet = quiet
+  do.call(
+    safe_install_packages,
+    c(list(
+        packages,
+        repos = repos,
+        type = type,
+        dependencies = dependencies,
+        quiet = quiet
+      ),
+      args
+    )
   )
 }
 
--- a/R/deps.R
+++ b/R/deps.R
@@ -111,7 +111,7 @@ local_package_deps <- function(pkgdir = ".", dependencies = NA) {
 
 dev_package_deps <- function(pkgdir = ".", dependencies = NA,
                              repos = getOption("repos"),
-                             type = getOption("pkgType")) {
+                             type = getOption("pkgType"), ...) {
 
   pkg <- load_pkg_description(pkgdir)
   repos <- c(repos, parse_additional_repositories(pkg))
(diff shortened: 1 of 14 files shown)


R/deps.R
```diff
@@ -280,25 +280,39 @@ update.package_deps <- function(object, ..., quiet = FALSE, upgrade = TRUE) {
...
 install_packages <- function(packages, repos = getOption("repos"),
                              type = getOption("pkgType"), ...,
-                             dependencies = FALSE, quiet = NULL,
-                             # These are options to `install()` used when
-                             # installing remotes, but can get passed to us by
-                             # `...` so we just ignore them here
-                             build, build_opts) {
+                             dependencies = FALSE, quiet = NULL) {
+
+  # We want to pass only args that exist in the downstream functions
```
jimhester: I don't really love this, but I am not sure of a better way to do it, you could list even more arguments here to catch them, but that seems like it would be a pain to maintain as the possible number of arguments in the `install_()` functions increases.

jimhester:
@gaborcsardi This could be reviewed now when you have time

jimhester:
Thanks for the review!

R/deps.R
hadley: Could we required that all install functions take `...` and then use ellipsis to warn if an argument is never evaluated?

R/deps.R
```diff
@@ -280,25 +280,39 @@ update.package_deps <- function(object, ..., quiet = FALSE, upgrade = TRUE) {
...
-    ...,
-    dependencies = dependencies,
-    quiet = quiet
+  do.call(
+    safe_install_packages,
+    c(list(
+        packages,
+        repos = repos,
+        type = type,
+        dependencies = dependencies,
```
hadley: Do we need to translate `dependencies = TRUE` to `dependencies = NA` to avoid installing all suggested dependencies of dependencies?

R/deps.R
gaborcsardi: remotes does not have dependencies, by design. ellipsis already has compiled code, and will trigger the DLL locking issue on Windows.

R/deps.R
gaborcsardi: Based on an empirical test, this works fine, at least `install_github("r-lib/processx", dependencies = TRUE)` only installs `Suggests` for processx.

R/deps.R
hadley: Oh, oops, I forgot about that constraint.

PR #146, opened 2018-08-28
Use credentials in install_git when using git2r

### LINKED ISSUE #106: Credentials in install_git

jimhester:
Fixes #106


jimhester:
I think the main question with this is the behavior if you try to use `credentials` with an non git2r backend. Right now it errors in this case.

jimhester:
Thanks for the review!

PR #148, opened 2018-08-30
Add `REMOTES_NO_EXTERNAL` environment variable

### LINKED ISSUE #147: Global option to avoid any dependencies (?)
A known problem on Windows, is that if a package with a shared library is loaded, the library DLL is  locked, and it cannot be overwritten. `R CMD INSTALL` still succeeds (with an extra message), and all other package files are installed, but the DLL is not, that belongs to the old version of the package.

This is even a problem in a clean session, if some (optional) recursive dependency or remotes is being installed/updated, one might end up with a broken installation.

If we could have a global option or env var to suppress loading optional dependencies, then we could handle all of these failures. (Updating remotes itself is fine, because it has no compiled code.)

jimhester:
To control use of external packages

Fixes #147

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,10 @@
   
 * `install_()` functions now pass arguments, including authentication
   information and upgrade down to dependencies (#53, #86, #87).
+* Remotes can be forced to use only its internal code by setting the
+  environment variable `R_REMOTES_STANDALONE` = "true". This is useful when
+  installing optional dependencies of remotes on Windows, such as curl or git2r
+  (#147)
 
 * `install_()` functions now return the name of the package(s) which were
   installed (#55).
--- a/R/download.R
+++ b/R/download.R
@@ -66,7 +66,7 @@ download_method <- function() {
 
 curl_download <- function(url, path, quiet) {
 
-  if (!pkg_installed("curl")) {
+  if (pkg_installed("curl")) {
     stop("The 'curl' package is required if R is older than 3.2.0")
   }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,18 +4,22 @@ sudo: false
 cache: packages
 latex: false
 
-r:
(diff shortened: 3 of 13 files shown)


jimhester:
I think the main question is what do we want to call the environment variable; I used `REMOTES_NO_EXTERNAL` and where do we document it?

NEWS.md
```diff
@@ -5,6 +5,11 @@
 
 # Development
 
+* Remotes can be forced to use only its internal code by setting the
+  environment variable `REMOTES_NO_EXTERNAL` = "true". This is useful when
```
gaborcsardi: I think the name is somewhat misleading, because "external" often refers in compiled code or system programs in R. Something like `R_REMOTES_STANDALONE` or  `R_REMOTES_NO_DEPS` or something like that would be better, I think.

R/download.R
```diff
@@ -66,7 +66,7 @@ download_method <- function() {
 
 curl_download <- function(url, path, quiet) {
 
-  if (!pkg_installed("curl")) {
+  if (!(use_external() && pkg_installed("curl"))) {
```
gaborcsardi: I am not sure this is needed here. We only call `curl_download()` if  `use_external()`, right?

gaborcsardi:
Re documenting it, maybe mention it in the README, and then add some more info to the package docs?

R/download.R
jimhester: You are probably right

jimhester:
Re: testing I am pretty sure we have tests for all code paths, both with and without the optional packages; that is part of the reason I have the `use_curl` etc arguments, so we can easily toggle them when testing.

NEWS.md
jimhester: `R_REMOTES_STANDALONE` sounds good to me

jimhester:
Thanks!

PR #149, opened 2018-09-05
Use pkg_installed() instead of is_installed()

joshuaulrich:
Commit f986e72b3574d8f9c7a5803a8d991fb370963e42 removed is_installed()
but did not replace all its calls with a suitable alternative.

See #147, #148.

### DIFF
--- a/install-github.R
+++ b/install-github.R
@@ -955,7 +955,7 @@ github_GET <- function(path, ..., host = "api.github.com", pat = github_pat()) {
 }
 
 github_commit <- function(username, repo, ref = "master",
-  host = "api.github.com", pat = github_pat(), use_curl = is_installed("curl"), current_sha = NULL) {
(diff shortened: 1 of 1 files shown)


jimhester:
I think this was a cause of two parallel PRs, one of which fixed removed the function, but other undid some of the uses.

Thanks for the PR!

joshuaulrich:
Thanks for the quick merge!  For what it's worth, I was motivated to create this PR because commit f986e72 broke r-hub while I was using it to track down a bug on an OS I don't have regular access to.

It was the second time a change in remotes had broken r-hub in ~10 days.  I expected r-hub to be more stable than that, so I suggested a solution in https://github.com/r-hub/rhub/issues/147.  Anyway, I just wanted to explain the motivation, in case you weren't aware.

Issue #34, opened 2016-08-26
Installation with dependencies should first download, then install


krlmlr:
I think installation of packages + dependencies should be split in three stages:
1. Download all source packages, including their dependencies
2. Install them in the correct order
3. Apply devtools-specific post-installation steps

For 2., we can leverage `install.packages()` with a `repos` argument that points to a local path: All we need to do is to download the packages to some `contrib/src/` subdirectory and call `tools::write_PACKAGES()`. This would finally give us parallel installation with `options(Ncpus > 1)`.

It looks like this would be relatively easy to implement with the current codebase of this repository.

Motivation: `remotes` does _a lot_ when installing packages and their dependencies, especially if these packages are from different remotes (e.g., CRAN + Bioconductor + GitHub). When revdep-checking RSQLite (with `bioconductor = TRUE`), I observed twice that I need 2-3 passes to install all revdeps. (The process is single-threaded and takes a day, I'm not too keen to repeat it _again_.)

Originally filed at https://github.com/hadley/devtools/issues/1308.

gaborcsardi:
One complication with this approach is the `Remotes` field. You would need to extract the package, to check `DESCRIPTION` in case there is a `Remotes` field there. And then `Remotes` fields can trigger new downloads again.

krlmlr:
Yes, that's right. But you download each package only once anyway, and the downloads can be cached easily, even persisting across sessions (=local CRAN-like repo).

gaborcsardi:
Sure. It is entirely possible, just not *that* trivial. `

gaborcsardi:
Some thoughts: 

* It would be great to have parallel downloads as well, and that is not entirely trivial. We would need to use `parallel` manually, and even the API that is not exported.
* It would be nice to have a command that does not download anything, but it gives you the dependency tree.
* Can we make the `Remotes` field easier to handle? E.g. we can put it in a DB with the commit hash included, and then if it is in the DB, we don't need to download the package to look at its `Remotes` field.

Ideally, we would do the downloads in parallel, and for non-CRAN-like packages we would extract `DESCRIPTION` as soon as the package is downloaded, to see if `Remotes` is there.

krlmlr:
Can futures help? I wouldn't want to overcomplicate things with parallel downloads, the last mile is often the bottleneck anyway.

Which package wins in a diamond dependency structure ABD ACD if only B but not C asks to install D from a specific remote?

gaborcsardi:
[...]

PR #150, opened 2018-09-05
Proof of concept for using curl with multiple requests

jimhester:
As a partial solution to #34 we could download all CRAN dependencies in parallel, then install them; we already have the full list of CRAN dependencies when we update them. This doesn't do anything for Remotes dependencies, but maybe that is ok.

As a simple test if this is worthwhile I used `download.packages()` to download binaries for all the dplyr CRAN dependencies, which ends up being 69 packages on my machine and the same thing with `download_packages()`

``` r
devtools::load_all("~/p/remotes")
#> Loading remotes
deps <- remotes::package_deps("dplyr", dependencies = TRUE)
deps <- deps[deps$is_cran, ]

out <- tempfile()
dir.create(out)
system.time(download.packages(deps$package, out, type = "binary"))
#>    user  system elapsed 
#>  10.897   1.404  59.585

out2 <- tempfile()
dir.create(out2)
system.time(download_packages(deps$package, out2, versions = deps$available, type = "binary"))
#> downloaded bindr_0.1.1.tgz:200
#> downloaded base64enc_0.1-3.tgz:200
#> downloaded assertthat_0.2.0.tgz:200
# (sic)
#>    user  system elapsed 
#>   0.437   0.766   6.024

expect_equal(list.files(out), list.files(out2))
```

Created on 2018-09-05 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

So at least in this scenario there is a substantial savings from the async downloads.

However there are still some things to be addressed including

1. `download.packages(type = "binary")` downloads the source package, so we would need to replicate code in `install.packages()` that determines whether a binary package is available.
2. Local `file:://` repositories are not currently addressed.
3. The current implementation does not handle all all the package types, it is hardcoded for mac binaries.
4. We would need to fallback to `download.packages()` if curl is not available.
5. We lose the individual progress bars while a download is running.

So @gaborcsardi my question is do you think it is worth the investment to fix the above deficiencies?

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -269,8 +269,11 @@ update.package_deps <- function(object, ..., quiet = FALSE, upgrade = TRUE) {
   }
 
   if (any(object$is_cran & behind)) {
-    install_packages(object$package[object$is_cran & behind], repos = attr(object, "repos"),
-      type = attr(object, "type"), ...)
+    to_install <- object$is_cran & behind
+    dest <- tempfile()
+    dir.create(dest)
+    res <- download_packages(object$package[to_install], dest, repos = attr(object, "repos"), type = "binary", versions = object$available[to_install])
+    install_packages(res[, 2], repos = NULL, type = "binary")
   }
 
   install_remotes(object$remote[!object$is_cran & behind], ..., quiet = quiet, upgrade = upgrade)
--- a/R/download.R
+++ b/R/download.R
@@ -109,3 +109,27 @@ download_method_secure <- function() {
     FALSE
   }
 }
+
+download_packages <- function (pkgs, destdir, repos = getOption("repos"),
+    contriburl = utils::contrib.url(repos, type), type = getOption("pkgType"), versions, ...) {
+
+  pool <- curl::new_pool()
+  on.exit(lapply(curl::multi_list(pool), curl::multi_cancel), add = TRUE)
+
+  save_package <- function(res) {
+    pkg <- basename(res$url)
+    if (res$status_code > 300) {
+      stop("Request failed: ", res$url, ": ", rawToChar(res$content), call. = FALSE)
+    }
+    cat("downloaded ", pkg, ":", res$status_code, "\n", sep = "")
+    writeBin(res$content, file.path(destdir, pkg))
+  }
+
+  pkg_url <- file.path(contriburl, paste0(pkgs, "_", versions, ".tgz"))
+  for (p in pkg_url) {
+    curl::curl_fetch_multi(p, save_package, pool = pool)
+  }
+  curl::multi_run(pool = pool)
+
+  matrix(ncol = 2, c(pkgs, file.path(destdir, basename(pkg_url))))
+}


gaborcsardi:
It feels like this will be cumbersome to implement correctly, so yeah, I am not sure, either.

Btw. `download.file()` can  download multiple files at once, if libcurl is supported. Maybe that's a better option for remotes:

```r
❯ system.time(download.file(c("https://httpbin.org/delay/3", "httpbin.org/delay/2"), destfile=c(t1 <- tempfile(), t2  <- tempfile()),  method  = "libcurl"))
trying URL 'https://httpbin.org/delay/3'
trying URL 'httpbin.org/delay/2'
   user  system elapsed
  0.006   0.004   4.127

❯ readLines(t1)
 [1] "{"
 [2] "  \"args\": {}, "
 [3] "  \"data\": \"\", "
 [4] "  \"files\": {}, "
 [5] "  \"form\": {}, "
 [6] "  \"headers\": {"
 [7] "    \"Accept\": \"*/*\", "
 [8] "    \"Connection\": \"close\", "
 [9] "    \"Host\": \"httpbin.org\", "
[10] "    \"User-Agent\": \"R (3.5.0 x86_64-apple-darwin15.6.0 x86_64 darwin15.6.0)\""
[11] "  }, "
[12] "  \"origin\": \"213.205.198.162\", "
[13] "  \"url\": \"https://httpbin.org/delay/3\""
[14] "}"

✔ 62.9 MiB
❯ readLines(t2)
 [1] "{"
 [2] "  \"args\": {}, "
 [3] "  \"data\": \"\", "
 [4] "  \"files\": {}, "
 [5] "  \"form\": {}, "
 [6] "  \"headers\": {"
 [7] "    \"Accept\": \"*/*\", "
 [8] "    \"Cache-Control\": \"max-age=259200\", "
 [9] "    \"Connection\": \"close\", "
[10] "    \"Host\": \"httpbin.org\", "
[11] "    \"User-Agent\": \"R (3.5.0 x86_64-apple-darwin15.6.0 x86_64 darwin15.6.0)\""
[12] "  }, "
[13] "  \"origin\": \"213.205.198.162\", "
[14] "  \"url\": \"http://httpbin.org/delay/2\""
[15] "}"
```

jimhester:
Yeah we could use `download.file(method = "libcurl")`, but I would unfortunately still have to do all of the above things either way. It also only issues a warning if a download fails which is kind of unfortunate.

jimhester:
Timing for `download.file(method = "libcurl")`, it is largely the same as for using the curl package.
```r
system.time(download.file(urls, outs, method = "libcurl"))
#> trying URL 'https://cloud.r-project.org/bin/macosx/el-capitan/contrib/3.5/assertthat_0.2.0.tgz'
#> trying URL 'https://cloud.r-project.org/bin/macosx/el-capitan/contrib/3.5/backports_1.1.2.tgz'
#> trying URL 'https://cloud.r-project.org/bin/macosx/el-capitan/contrib/3.5/base64enc_0.1-3.tgz'
#> trying URL 'https://cloud.r-project.org/bin/macosx/el-capitan/contrib/3.5/BH_1.66.0-1.tgz'
#> trying URL 'https://cloud.r-project.org/bin/macosx/el-capitan/contrib/3.5/bindr_0.1.1.tgz'
#> (sic)
#>    user  system elapsed
#>   0.636   1.434   7.461
```

PR #151, opened 2018-09-07
fix internal remotes:::untar wrapper for Windows platforms

bleutner:
utils::untar returns a warning when it fails but the tryCatch in remotes:::untar expects an error. 
Therefore the alternative untar without  extras="--force-local" is not triggered when it should be and the installation from a tarball fails.

tryCatch was modified to expect a warning instead.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -157,10 +157,10 @@ with_rprofile_user <- function(new, code) {
 
 untar <- function(tarfile, ...) {
   if (os_type() == "windows") {
-    tryCatch(
-      utils::untar(tarfile, extras = "--force-local", ...),
-      warning = function(e) utils::untar(tarfile, ...)
-    )
+    status <- try(utils::untar(tarfile, extras = "--force-local", ...))
+    if(inherits(status, "try-error") || status != 0){
+        utils::untar(tarfile, ...)
+    }
 
   } else {
     utils::untar(tarfile, ...)
--- a/R/utils.R
+++ b/R/utils.R
@@ -159,7 +159,7 @@ untar <- function(tarfile, ...) {
   if (os_type() == "windows") {
     tryCatch(
       utils::untar(tarfile, extras = "--force-local", ...),
-      error = function(e) utils::untar(tarfile, ...)
+      warning = function(e) utils::untar(tarfile, ...)
     )
 
   } else {
--- a/R/utils.R
+++ b/R/utils.R
@@ -160,8 +160,9 @@ untar <- function(tarfile, ...) {
     status <- try(utils::untar(tarfile, extras = "--force-local", ...))
     if(inherits(status, "try-error") || status != 0){
         utils::untar(tarfile, ...)
+    } else {
+        status
     }
-
   } else {
     utils::untar(tarfile, ...)
   }


gaborcsardi:
Thanks! I see some issues with this. 

* `untar()` may actually return an error, e.g. when `untar2()` is used. So we need to keep the `error()` branch as well.
* Other warnings, e.g. https://github.com/wch/r-source/blame/5b2f6807839f2970051f6723040a27bb97a49c81/src/library/utils/R/tar.R#L51 may also happen, and we don't want an _exiting_ handlers for these.

I think maybe it would be better to 
* keep the `error`, and
* handle the return value of `utils::untar`, this is non-zero if an error happens, and this behavior  hasn't changed for a long time: https://github.com/wch/r-source/blame/5b2f6807839f2970051f6723040a27bb97a49c81/src/library/utils/R/tar.R#L113-L115

jimhester:
Related to this, tar on windows has issues with symlinks, e.g. installing remotes on windows with the tar bundled with RTools.

```
> remotes::install_github("r-lib/remotes", force=T)
Downloading GitHub repo r-lib/remotes@master
/3.5/bin/tar: r-lib-remotes-fb6682b/README.markdown: Cannot create symlink to 'inst/README.markdown': No such file or directory
/3.5/bin/tar: Exiting with failure status due to previous errors
Warning in utils::untar(tarfile, extras = "--force-local", ...) :
  ‘tar.exe -zxf "C:\Users\IEUser\AppData\Local\Temp\Rtmpu2t9O5\file192ce5a3d6e.tar.gz" -C "C:/Users/IEUser/AppData/Local/Temp/Rtmpu2t9O5/remotes192c689f4713" --force-local’ returned error code 2
```

And forcing to use the internal tar
```
> Sys.setenv("TAR" = "internal")
> remotes::install_github("r-lib/remotes", force=T)
Warning in untar2(tarfile, files, list, exdir, restore_times) :
  failed to copy ‘r-lib-remotes-fb6682b/inst/README.markdown’ to ‘r-lib-remotes-fb6682b/README.markdown’
Warning in untar2(tarfile, files, list, exdir, restore_times) :
  skipping pax global extended headers
Warning in untar2(tarfile, files, list, exdir, restore_times) :
  skipping pax global extended headers
```

The warnings due to pax global headers are expected and OK. But for the symlinks I am not really sure what we can do about this.

gaborcsardi:
Can we call external tar  with `-h`?

gaborcsardi:
Also, the internal tar is kind of good in this case. The R package itself should not contain symlinks, anyway.

jimhester:
`-h` is only for _creating_ tar archives, not for extracting.

From https://github.com/heroku/heroku-slugs/issues/3 the only real solution seems to be running the untar command multiple times; as we do not control creating of the tarballs (GitHub does). Or just ignoring this error. Alternatively we could remove the symlinks from the packages, what is the reason you symlink NEWS.md and README.markdown?

> It appears that GNU tar (for at least the MSYS2 package of it) does partially handle filesystems that don't support symlinks, in that whilst extracting it does replace the symlink with a copy of the full file, if the file already exists on the filesystem.
>
> The problem with this is that it's dependant on the order of the files in the tar, since if the symlink is extracted first, the target won't yet exist (this isn't a problem on linux since dangling symlinks are allowed). This comes from the fact that tar is intended as a streaming format, though perhaps GNU tar should still try and do the right thing when in seekable mode (but fixing GNU tar is beyond the scope of this ticket).
>
> As such, a workaround is to run the tar extraction multiple times, which will eventually succeed without errors once all the symlink targets have been created

jimhester:
I guess we could just force `tar = "internal"`, then we get individual warnings for this we can detect and retry

gaborcsardi:
Yeah, internal tar is really slow, though. :(

gaborcsardi:
But most packages don't have symlinks....

bleutner:
not sure what to do about the symlinks issue. 

For the error handling I've switched to try() instead of tryCatch(), which should now handle both regular errors and errors communicated via the exit value.
Would that work?

R/utils.R
```diff
@@ -157,10 +157,10 @@ with_rprofile_user <- function(new, code) {
...
 untar <- function(tarfile, ...) {
   if (os_type() == "windows") {
-    tryCatch(
-      utils::untar(tarfile, extras = "--force-local", ...),
-      error = function(e) utils::untar(tarfile, ...)
-    )
+    status <- try(utils::untar(tarfile, extras = "--force-local", ...))
+    if(inherits(status, "try-error") || status != 0){
+        utils::untar(tarfile, ...)
+    }
```
gaborcsardi: I would put  an 
```
} else { 
  status 
}
```  
here, just  to always return  the same thing, and then it looks good!

gaborcsardi:
@jimhester  I think this is good to merge. The CI failures are unrelated.

PR #155, opened 2018-09-11
Expose the full GitHub error messages

jimhester:
If we are using curl we can get the full error message from the JSON
response.

This is particularly useful for 403 errors when you have exhausted the
rate limit.

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -1,12 +1,29 @@
 
-github_GET <- function(path, ..., host = "api.github.com", pat = github_pat()) {
+github_GET <- function(path, ..., host = "api.github.com", pat = github_pat(), use_curl = !is_standalone() && pkg_installed("curl")) {
 
   url <- build_url(host, path)
 
-  tmp <- tempfile()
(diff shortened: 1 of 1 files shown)


R/github.R
```diff
@@ -93,3 +112,8 @@ github_DESCRIPTION <- function(username, repo, subdir = NULL, ref = "master", ho
     base64_decode(gsub("\\\\n", "", fromJSONFile(tmp)$content))
   }
 }
+
+github_error_message <- function(res) {
+  msg <- fromJSON(rawToChar(res$content))
```
gaborcsardi: I would put  `simplifyVector  = FALSE` here, to be on the safe side.

jimhester: This is the internal fromJSON from https://github.com/r-lib/remotes/blob/fb6682b6091412e75e313345ab54e96e82ed58b7/R/json.R#L31, so I guess it never simplifies.

gaborcsardi: Oh, right, I wrote that, I remember now.  :)

PR #156, opened 2018-09-12
Expose all installation options to all install functions

jimhester:
This should make things more clear what is available, all options should
be now exposed and documented other than those to `install.packages()`

Mainly to address the last bit of https://github.com/r-lib/remotes/issues/129#issuecomment-420636312

> Just a small note that upgrade isn't a documented parameter of remotes::install_github(), nor is it a documented parameter of install.packages() (which is where the documentation of install_github says the ... are passed).

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -23,6 +23,7 @@
 #'
 #' @param object A `package_deps` object.
 #' @param ... Additional arguments passed to `install_packages`.
+#' @inheritParams install_github
 #'
 #' @return
 #'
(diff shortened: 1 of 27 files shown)


gaborcsardi:
Hmmm, but then of course the default values are scattered all over, and it will be hard to change them. It  would be nice to have a general solution to this, e.g. roxygen could (dynamically?) generate a section for the args that can be passed down.

Hopefully we are not creating "matches multiple arguments" errors....

jimhester:
Yes, I agree it has downsides and was a pain to add them everywhere. It would be nice to be able to have roxygen / something else forward arguments. We experimented with this a bit in https://github.com/tidyverse/ggplot2/pull/1763

I deliberately removed the default arguments to the internal functions so that we would catch multiple arguments / missing arguments, and caught a few when writing this, so I think it is ok 🤞

PR #157, opened 2018-09-12
Put RTools on the path when necessary

jimhester:
This avoids users needing to manually add the RTools location to their
PATH; which is error prone as you then have to update it for newer
RTools versions and ensure the unix tools bundled with RTools don't
conflict with other tools on your PATH.

This will require a pkgbuild release, as `local_build_tools()` does not exist in the CRAN version.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,12 +26,14 @@ Suggests:
     covr,
     git2r (>= 0.23.0),
     mockery,
-    pkgbuild,
+    pkgbuild (>= 1.0.0.9000),
     pingr,
     testthat,
     withr
 Depends:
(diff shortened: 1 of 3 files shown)


jimhester:
So this will require a small pkgbuild release before we can release remotes, but I think it is worth having as this will make adding rtools to the PATH transparent to the user (assuming you have pkgbuild installed).

jimhester:
I think it is ok to just have the functionality be optional. Users will typically have the full devtools stack, so will have pkgbuild, but on travis / appveyor etc. they won't and this will fall back to the previous warning instead (which is fine), as both of those cases have Rtools on the PATH (or don't use rtools at all)

PR #168, opened 2018-09-24
Rename branch to ref

jimhester:
This makes it more clear what values the argument can take, it can be
any git reference, not just a branch


R/install-remote.R
```diff
@@ -165,13 +165,13 @@ package2remote <- function(name, lib = .libPaths(), repos = getOption("repos"),
       sha = x$RemoteSha),
     xgit = remote("xgit",
       url = trim_ws(x$RemoteUrl),
-      branch = x$RemoteBranch,
+      ref = x$RemoteRef %||% x$RemoteBranch,
```
jimhester: I at least have a fallback to RemoteBranch here, so previously installed packages will still work. We can retain the RemoteBranch as well though I guess until the branch is fully removed.

PR #169, opened 2018-09-24
Only install remotes that are in the stated CRAN dependencies

### LINKED ISSUE #167: Remotes installation too aggressive?
We talked about this before, probably, but maybe we should not install `Remotes` for packages in `Suggests`, if suggested packages are not requested otherwise?

At least it is quite strange that even if I set `dependencies = FALSE`, `install_github("r-lib/remotes")` still installs pkgbuild, because it is in `Remotes`.

One issue here is that we don't necessarily know the name of a remote, but I think we already handle that somewhere by requiring the new remote syntax to specify the package name explicitly, if it differs from the repo (or subdir?) name?

jimhester:
If we filter the dependencies like this we will ensure only remote dependencies listed in the CRAN dependencies will be installed. By this point we have already included only CRAN dependencies requested, e.g. if they are in Suggests and `dependencies = NA` they won't be here.

I think it used to work like this and the reason we went to always installing remotes is some people don't put the remote package in one of the other fields. But I agree it is nice to not install a remote if it is only in Suggests (and you are not installing Suggests), so I am ok changing the behavior.

Fixes #167

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -134,14 +134,13 @@ dev_package_deps <- function(pkgdir = ".", dependencies = NA,
 }
 
 combine_deps <- function(cran_deps, remote_deps) {
-  deps <- rbind(cran_deps, remote_deps)
-
   # Only keep the remotes that are specified in the cran_deps
-  # Keep only the Non-CRAN remotes if there are duplicates as we want to install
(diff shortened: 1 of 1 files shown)


gaborcsardi:
Thanks!

PR #172, opened 2018-09-27
Fixing untar on Windows

### LINKED ISSUE #171: tar.exe: Option --force-local is not supported
When I use `remotes::install_local(force = TRUE)`, because of "Skipping install of '...' from a local remote, the SHA1 (0.1.0-90) has not changed since last install." I get the error `tar.exe: Option --force-local is not supported`.
I use a Windows 10 system.

Here is the complete output after the "DONE" statement:

    * DONE (...)                                                                                                                                     
    tar.exe: Option --force-local is not supported                                                                                                            
    Usage:                                                                                                                                                    
      List:    tar.exe -tf <archive-filename>                                                                                                                 
      Extract: tar.exe -xf <archive-filename>                                                                                                                 
      Create:  tar.exe -cf <archive-filename> [filenames...]                                                                                                  
      Help:    tar.exe --help                                                                                                                                 
    Error in if (inherits(status, "try-error") || status != 0) { :                                                                                            
      missing value where TRUE/FALSE needed                                                                                                                   
    In addition: Warning message:                                                                                                                             
    In system(cmd, intern = TRUE) :                                                                                                                           
      running command 'tar.exe -ztf "C:\Users\Name\AppData\Local\Temp\RtmpScq6Pl\file2688718c77bb\..._0.1.0-9000.tar.gz" --force-local' had status 1

Do you have an idea?

gaborcsardi:
`utils::untar()` calls `system()` with `intern = TRUE`,
so we need to check for `attr(status, "status")` as well.
In addition we silence the first `try()`, and give an
explicit message about retrying instead. See also #151.

Closes #171.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -166,9 +166,7 @@ untar <- function(tarfile, ...) {
       status <- try(
         suppressWarnings(utils::untar(tarfile, extras = "--force-local", ...)),
         silent = TRUE)
-      if (! inherits(status, "try-error") &&
-          ! is_error_status(status) &&
-          !is_error_status(attr(status, "status"))) {
+      if (! is_tar_error(status)) {
         return(status)
 
       } else {
@@ -178,6 +176,12 @@ untar <- function(tarfile, ...) {
   }
 
   utils::untar(tarfile, ...)
+}
+
+is_tar_error <- function(status) {
+  inherits(status, "try-error") ||
+    is_error_status(status) ||
+    is_error_status(attr(status, "status"))
 }
 
 is_error_status <- function(x) {
--- a/R/utils.R
+++ b/R/utils.R
@@ -157,27 +157,27 @@ with_rprofile_user <- function(new, code) {
 
 untar <- function(tarfile, ...) {
   if (os_type() == "windows") {
+
     tarhelp <- tryCatch(
       system2("tar", "--help", stdout = TRUE, stderr = TRUE),
       error = function(x) "")
-    forcelocal <- any(grepl("--force-local", tarhelp))
-    if (forcelocal)  {
+
+    if (any(grepl("--force-local", tarhelp)))  {
       status <- try(
         suppressWarnings(utils::untar(tarfile, extras = "--force-local", ...)),
         silent = TRUE)
-      if (inherits(status, "try-error") ||
-          is_error_status(status) || is_error_status(attr(status, "status"))) {
-        message("External tar failed with `--force-local`, trying without")
-        utils::untar(tarfile, ...)
+      if (! inherits(status, "try-error") &&
+          ! is_error_status(status) &&
+          !is_error_status(attr(status, "status"))) {
+        return(status)
+
       } else {
-        status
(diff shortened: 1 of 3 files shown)


gaborcsardi:
It seems like I cannot  suppress the output from `untar()`.  I'll try a different approach and call  `tar --help` first, and see  if `--force-local` is mentioned there.

gaborcsardi:
@jimhester I think this is good now.

R/utils.R
```diff
@@ -157,17 +157,31 @@ with_rprofile_user <- function(new, code) {
 
 untar <- function(tarfile, ...) {
   if (os_type() == "windows") {
-    status <- try(utils::untar(tarfile, extras = "--force-local", ...))
-    if(inherits(status, "try-error") || status != 0){
+    tarhelp <- system2("tar", "--help", stdout = TRUE, stderr = TRUE)
```
jimhester: If `Sys.getenv("TAR")` is 'internal' we should probably just skip this completely. Also what does `system2()` do if there is no tar executable on the PATH?

OTOH `extras` is ignored if using the internal tar implementation IIRC, so maybe it is fine as is.

gaborcsardi: Oh, yeah, `system2` needs to  be `tryCatch()`-ed. Thanks for catching this.

R/utils.R
```diff
@@ -157,17 +157,31 @@ with_rprofile_user <- function(new, code) {
...
+    forcelocal <- any(grepl("--force-local", tarhelp))
+    if (forcelocal)  {
+      status <- try(
+        suppressWarnings(utils::untar(tarfile, extras = "--force-local", ...)),
+        silent = TRUE)
+      if (inherits(status, "try-error") ||
+          is_error_status(status) || is_error_status(attr(status, "status"))) {
+        message("External tar failed with `--force-local`, trying without")
         utils::untar(tarfile, ...)
```
jimhester: There are three cases that all end with `utils::untar(tarfile, ...)`. So we should be able to rewrite the case that returns the status with an early return and have the rest of the cases fall through.

gaborcsardi: I tried that but it was actually less readable than the current code. You cannot mark "not done  yet"  with `status =  NULL`, because maybe that's  a legit return value. Do you need a `done` variable, plus `status` ... anyway, I'll give it another try.

R/utils.R
```diff
@@ -157,15 +157,31 @@ with_rprofile_user <- function(new, code) {
...
+      system2("tar", "--help", stdout = TRUE, stderr = TRUE),
+      error = function(x) "")
+
+    if (any(grepl("--force-local", tarhelp)))  {
+      status <- try(
+        suppressWarnings(utils::untar(tarfile, extras = "--force-local", ...)),
+        silent = TRUE)
+      if (! inherits(status, "try-error") &&
+          ! is_error_status(status) &&
+          !is_error_status(attr(status, "status"))) {
```
jimhester: Could we just roll these all into `is_error_status()`?

gaborcsardi:
Thanks!

PR #176, opened 2018-10-02
install_gitlab fails bc it wants the phrase "private_token" (take 2)

overmar:
I reverted the spaces and added commas, sorry that this is another pull, RStudio and I were having issues with the spaces being eaten.


overmar:
However I now wonder if download should be method classed, as bitbucket api v2 isn't going to follow the normal download conventions either.

gaborcsardi:
Thanks!

PR #178, opened 2018-10-02
Helpful message when build fails for long paths

### LINKED ISSUE #84: install_github() failing on Windows due to long path names
<a href="https://github.com/MrDAndersen"></img></a> **Issue by [MrDAndersen](https://github.com/MrDAndersen)**
_Thursday Jun 23, 2016 at 16:15 GMT_
_Originally opened as https://github.com/hadley/devtools/issues/1235_

----

We have a package available for install in a sub directory `R Package/ffanalytics` at `dadrivr/FantasyFootballAnalyticsR`. Users are instructed to install this via `devtools::install_github(repo = "dadrivr/FantasyFootballAnalyticsR", subdir = "R Package/ffanalytics")`. However, Windows users are reporting that the install fails with following error

```
Error in utils::unzip(src, exdir = target) :
cannot open file ‘C:/Users/<username>/AppData/Local/Temp/RtmpMZfFU0/devtools63c36d21285/dadrivr-FantasyFootballAnalyticsR-6c6c4a9/R Markdown/SubcriptionAccuracy/subscriptionAccuracy_cache/markdown_phpextra+backtick_code_blocks/unnamed-chunk-1_af5e2a896358b9ccbc8241e876b9ce5c.RData’: No such file or directory
```

The file mentioned is not part of the package directory, so I suspect that `devtools::install_github` is downloading the entire repo and then parsing it to find the subdir with the package. Since the repo files are extracted into a temp directory (`C:/Users/<username>/AppData/Local/Temp......`) the file names get rather long when it has to go through several levels of folders. 
The interesting thing is that you can get the command to complete after a number of tries - which may be when the file paths become short enough (<256 characters)

gaborcsardi:
On windows. Closes #84.

This is for the case when we call `R CMD build` directly. Otherwise this should be "fixed" in pkgbuild.

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -91,13 +91,32 @@ safe_build_package <- function(pkgdir, build_opts, dest_path, quiet, use_pkgbuil
 
     pkgdir <- normalizePath(pkgdir)
 
+    message("Running `R CMD build`...")
     in_dir(dest_path, {
       with_envvar(env, {
-        output <- rcmd("build", c(build_opts, shQuote(pkgdir)), quiet = quiet)
+        output <- rcmd("build", c(build_opts, shQuote(pkgdir)), quiet = quiet,
+                       fail_on_status = FALSE)
       })
     })
 
-    file.path(dest_path, sub("^[*] building[^[:alnum:]]+([[:alnum:]_.]+)[^[:alnum:]]+$", "\\1", output[length(output)]))
+    if (output$status != 0) {
+      cat("STDOUT:\n")
+      cat(output$stdout, sep = "\n")
+      cat("STDERR:\n")
+      cat(output$stderr, sep = "\n")
+      if (sys_type() == "windows" &&
+          any(grepl("over-long path length", output$stderr))) {
+        message(
+          "\nIt seems that this package contains files with very long paths.\n",
+          "This is not supported on most Windows versions. Please contant the\n",
+          "package authors and tell them about this. See this GitHub issue\n",
+          "for more details: https://github.com/r-lib/remotes/issues/84\n")
+      }
+      stop(sprintf("Failed to `R CMD build` package, try `build = FALSE`."),
+           call. = FALSE)
+    }
+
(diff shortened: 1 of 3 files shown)


R/install.R
```diff
@@ -91,13 +91,32 @@ safe_build_package <- function(pkgdir, build_opts, dest_path, quiet, use_pkgbuil
...
+    if (output$status != 0) {
+      cat("STDOUT:\n")
+      cat(output$stdout, sep = "\n")
+      cat("STDERR:\n")
+      cat(output$stderr, sep = "\n")
+      if (sys_type() == "windows" &&
+          any(grepl("over-long path length", output$stderr))) {
+        message(
+          "\nIt seems that this package contains files with very long paths.\n",
+          "This is not supported on most Windows versions. Please contant the\n",
```
jimhester: contant -> contact

R/install.R
```diff
@@ -91,13 +91,32 @@ safe_build_package <- function(pkgdir, build_opts, dest_path, quiet, use_pkgbuil
...
       })
     })
 
-    file.path(dest_path, sub("^[*] building[^[:alnum:]]+([[:alnum:]_.]+)[^[:alnum:]]+$", "\\1", output[length(output)]))
+    if (output$status != 0) {
+      cat("STDOUT:\n")
+      cat(output$stdout, sep = "\n")
+      cat("STDERR:\n")
+      cat(output$stderr, sep = "\n")
+      if (sys_type() == "windows" &&
```
jimhester: Maybe put this check in a function `check_for_long_paths()` or similar?

R/install.R
gaborcsardi: done.

R/install.R
gaborcsardi: done.

gaborcsardi:
@jimhester done.

gaborcsardi:
@jimhester Thanks!

Issue #177, opened 2018-10-02
install_bitbucket ref in auth issue


overmar:
It appears that when it is a public repo there are no problems
```
remotes::install_bitbucket(repo = "roverman/remotes.test@dev")
remotes::install_bitbucket(repo = "roverman/remotes.test", ref = "dev")
```
But a private repo
```
remotes::install_bitbucket(repo = <repo>, ref = "dev", auth_user = <auth_user> password = <password>)
Error in utils::download.file(url, path, method = download_method(), quiet = quiet,  : 
  cannot open URL 'https://<auth_user>:<password>@api.bitbucket.org/2.0/repositories/<repo>/src/master/DESCRIPTION'
```
It appears that the download is not respecting the other path

PR #179, opened 2018-10-02
ref does not work with bitbucket auth

overmar:
Fixes https://github.com/r-lib/remotes/issues/177

### DIFF
--- a/R/install-bitbucket.R
+++ b/R/install-bitbucket.R
@@ -115,7 +115,8 @@ remote_package_name.bitbucket_remote <- function(remote, ...) {
 
   bitbucket_DESCRIPTION(
     username = remote$username, repo = remote$repo,
-    host = remote$host, auth = basic_auth(remote))$Package
+    host = remote$host, auth = basic_auth(remote),
+    ref = remote$ref)$Package
 }
 
 #' @export


R/install-bitbucket.R
```diff
@@ -115,7 +115,8 @@ remote_package_name.bitbucket_remote <- function(remote, ...) {
 
   bitbucket_DESCRIPTION(
     username = remote$username, repo = remote$repo,
-    host = remote$host, auth = basic_auth(remote))$Package
+    host = remote$host, auth = basic_auth(remote),
+    ref = remote$ref)$Package
```
gaborcsardi: Sg is wrong here...

gaborcsardi: Oh, no,  sorry my bad, forget  it....

overmar: No worries, I can't remember to use a <- or put spaces in after commas. :)

gaborcsardi:
I suppose you tried it and it works?

overmar:
It does, everything correctly reassociates. Apparently the only place it was having trouble was in the description, I think the path had been included in the rest of the calls that created the url. Also it works with both @ and ref = again.

gaborcsardi:
Thanks!

R/install-bitbucket.R
jimhester: We need the subdirectory here as well, and it would likely be easier to read this if the arguments were in the same order as the formals of `bitbucket_DESCRIPTION`.

```r
bitbucket_DESCRIPTION(
  username = remote$username, repo = remote$repo,
  subdir = remote$subdir, ref = remote$ref,
  host = remote$host, auth = basic_auth(remote)
)$Package
```

gaborcsardi: @jimhester OK, I can do that.

gaborcsardi: done.

PR #182, opened 2018-10-03
Warn if R.home() has space on windows, old R

### LINKED ISSUE #98: C:\Program is not recognized as internal or external program
<a href="https://github.com/pfgherardini"></img></a> **Issue by [pfgherardini](https://github.com/pfgherardini)**
_Monday May 22, 2017 at 22:41 GMT_
_Originally opened as https://github.com/hadley/devtools/issues/1514_

----

Hi All,

I realize this is probably a bug in R-core, but I figured maybe a workaround can be implemented in devtools, since it shows up because of the way R is invoked from this package.

Basically on Windows, depending on system configuration, `install_github` fails with a message similar to 
``C:\Program is not recognized as internal or external program``

This happens when both of the following conditions are met
- R is installed in a directory that contains whitespace in its name (e.g. "Program Files")
- Short filenames (SFN) are disabled for the volume on which R is installed

The second point can be verified with this command (assuming R is installed on the C drive; requires admin privileges, more info [here](https://technet.microsoft.com/en-us/library/cc785435(v=ws.11).aspx))

 ``fsutil.exe behavior query disable8dot3 C:``

if the output looks like

``
8dot3 name creation is disabled on C:
``

then you are in trouble because R will not be able to handle a path with spaces properly, AND it will not be able to get an SFN-style path because it's disabled on your volume.

A telltale sign of this problem is that you won't be able to start `R.exe` from the command line, even though the RGui will start fine. This does not create a problem for the built-in `install.packages` function because it does not spawn another R process like `devtools` seems to do.

gaborcsardi:
Closes #98.

Hopefully we won't have any false positives.

It looks like this:
```
> install_version("pkgconfig")
Downloading package from url: https://cloud.r-project.org/src/contrib/pkgconfig_2.0.2.tar.gz
Warning in warn_for_potential_errors() :
  
!!! Installation will probably fail!
This version of R has trouble with building and installing packages if
the R HOME directory (currently 'E:/Program Files/R/R-3.4.2')
has space characters. Possible workarounds include:
- installing R to the C: drive,
- installing it into a path without a space, or
- creating a drive letter for R HOME via the `subst` windows command, and
  starting R from the new drive.
See also https://github.com/r-lib/remotes/issues/98

Running "E:/Program Files/R/R-3.4.2/bin/x64/Rcmd.exe" build "C:\Users\Gabor\AppData\Local\Temp\RtmpeCW82i\remotes98c6abce2d\pkgconfig" --no-resave-data \
  --no-resave-data --no-manual --no-build-vignettes
'E:\Program' is not recognized as an internal or external command,
operable program or batch file.
Error in run(bin, args = real_cmdargs, stdout_line_callback = real_callback(stdout),  : 
  System command error
>
```

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -1,6 +1,8 @@
 install <- function(pkgdir, dependencies, quiet, build, build_opts, upgrade,
                     repos, type, ...) {
 
+  warn_for_potential_errors()
+
   if (file.exists(file.path(pkgdir, "src"))) {
     if (has_package("pkgbuild")) {
       pkgbuild::local_build_tools(required = TRUE)
--- a/R/utils.R
+++ b/R/utils.R
@@ -388,3 +388,19 @@ format_str <- function(x, width = Inf, trim = TRUE, justify = "none", ...) {
   }
   x
 }
+
+warn_for_potential_errors <- function() {
+  if (sys_type() == "windows" && grepl(" ", R.home()) &&
+      getRversion() <= "3.4.2") {
+    warning(immediate. = TRUE,
+      "\n!!! Installation will probably fail!\n",
+      "This version of R has trouble with building and installing packages if\n",
+      "the R HOME directory (currently '", R.home(), "')\n",
+      "has space characters. Possible workarounds include:\n",
+      "- installing R to the C: drive,\n",
+      "- installing it into a path without a space, or\n",
+      "- creating a drive letter for R HOME via the `subst` windows command, and\n",
+      "  starting R from the new drive.\n",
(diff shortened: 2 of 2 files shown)


jimhester:
I am not sure we should bother doing this especially if it is fixed in newer versions of R, but I guess the check is not too difficult.

gaborcsardi:
We can also add the workarounds to the issue and point people there. 

I think the warning is better, plenty of older R versions are still around, I am sure. We could keep it until 3.4.x goes out of support. That'll be a couple of years...

jimhester:
Ok, looks good, I will merge.

PR #185, opened 2018-10-08
Allow users to choose what to upgrade

jimhester:
So this is another stab at the upgrading, showing a menu when run interactively that allows users to choose which packages to upgrade.

It uses `utils::select.list()` for the menu, which may or may not be the best choice...


R/deps.R
```diff
@@ -17,7 +17,10 @@
 #'   and is the default. `FALSE` is shorthand for no dependencies (i.e.
 #'   just check this package, not its dependencies).
 #' @param quiet If `TRUE`, suppress output.
-#' @param upgrade If `TRUE`, also upgrade any of out date dependencies.
+#' @param upgrade If `NA` and interactive prompt to upgrade out of date
```
gaborcsardi: This is a bit hard to read I think, bc interactive could an adjective, as  in "interactive prompt". I had to read  it three times  to understand.

R/deps.R
```diff
@@ -236,14 +239,17 @@ UNAVAILABLE <- 2L
 
 update.package_deps <- function(object,
                            dependencies = NA,
-                           upgrade = TRUE,
+                           upgrade = NA,
```
gaborcsardi: How about having  `"ask"` instead of  `NA`? And the default could be `if  (interactive()) "ask" else TRUE`. That would allow us to have other "modes" later, e.g. `"cran-only"`, etc. and it is also more readable. In `install.packages()` I have to check the docs every single time when I see `dependencies = NA`  or `NULL` or  `TRUE`, etc.

jimhester: I used NA so it would take only logical inputs, but I suppose the distinction is minor.

gaborcsardi: Yeah, I think in languages with constants, this would be a constant or enum, which in R is basically a string. `NA` is  good to make it logical, but I think it is just cleaner to make it a "constant", that is also extendable.

jimhester: I mostly agree with you I think in this case, I posted on slack to see if anyone else has opinions.

jimhester:
I refactored this a bunch and added tests for all the paths, so taking another quick look would be appreciated!

gaborcsardi:
Looks pretty good!

jimhester:
Thanks for reviewing!

PR #191, opened 2018-10-10
Draft implementation of install_dev

jimhester:


jimhester:
Ok I reworked this a bunch, it now looks in both URL and BugReports fields and added a bunch of tests with packages with various edge cases.

I think the one edge case I don't have a test for is a package in a subdirectory, although I think the code is correct and will work in that case. I just don't know of any CRAN packages offhand that are installed from a subdirectory.

gaborcsardi:
feather? AFK, so not sure...

jimhester:
feather is a good example, but unfortunately won't work because the subdir is not included in the URL (https://github.com/wesm/feather). If it was listed as https://github.com/wesm/feather/R it would be perfect. :(

jimhester:
I think this is good to merge if it looks ok to you @gaborcsardi

R/install-dev.R
```diff
@@ -0,0 +1,75 @@
+#' Install the development version of a package
+#'
+#' `install_dev()` retrieves the package DESCRIPTION from the CRAN mirror and
+#' looks in the 'URL' and 'BugReports' fields for GitHub, GitLab or Bitbucket
+#' URLs. It then calls the appropriate `install_()` function to install the
+#' development package.
+#'
+#' @param package The package name to install.
+#' @param cran_url The URL of the CRAN mirror to use, if unset uses 'https://cloud.r-project.org'.
```
gaborcsardi: This is not actually true. I mean, maybe mention the `repos` option here.

PR #192, opened 2018-10-10
Add pkgdown site

jimhester:
I created the key with `travis::use_travis_deploy(travis_repo = "r-lib/remotes")`, this should be all the rest we need for the site. Assuming this works we can switch it to remotes.r-lib.org.

Deploys don't happen from PRs, so we would have to merge this to see if it works.


gaborcsardi:
Would it make sense to merge it to a non-master branch, and then refine it there?

PR #193, opened 2018-10-10
Pkgdown

jimhester:
Ok this should deploy, assuming it works before we merge we can remove the second commit (963f82b)


jimhester:
Note to self for tomorrow, need to add `install_dev()` to the reference.

PR #194, opened 2018-10-10
Include the message in the response

jimhester:
In both cases, an R condition object and the downloaded error page the
response will be in response$message, so we can use that
unconditionally.

This should expose the full error message explaining what has gone
wrong.

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -14,8 +14,7 @@ github_GET <- function(path, ..., host = "api.github.com", pat = github_pat(), u
     res <- curl::curl_fetch_memory(url, handle = h)
 
     if (res$status_code >= 300) {
-      stop("HTTP error ", res$status_code, ".",
-           "\n", github_error_message(res), call. = FALSE)
+      stop(github_error(res))
     }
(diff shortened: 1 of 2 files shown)


jimhester:
I am pretty sure this will work as-is, alternatively it might be useful to return some rate limit information, such as the rate limit remaining and when the rate limit will be reset.

jimhester:
Now looks like this

```r
github_GET("repos/foo/bar")
#> Error: HTTP error 404.
#>   Not Found
#> 
#>   Rate limit remaining: 4999
#>   Rate limit reset at: 2018-10-10 19:43:52 UTC
```

gaborcsardi:
Cool. It is a pity we only have this  for curl....

jimhester:
When I switch travis over to use remotes I was planning on always installing curl as well, so at least that case will have it...

gaborcsardi:
Yeah, it is fine, it is just annoying to have all these code paths.

jimhester:
Agreed!

jimhester:
This seems to be working well.

PR #198, opened 2018-10-11
Copy over remotes vignette from devtools

gaborcsardi:


vignettes/dependencies.Rmd
```diff
@@ -0,0 +1,91 @@
...
+A type of 'github' can be specified, but is not required
+
+```yaml
+Remotes: github::tidyverse/ggplot2
+```
+
+### Other sources
+
+All of the currently supported install sources are available, see the 'See
+Also' section in `?install` for a complete list.
```
jimhester: I think this needs to be updated, as there is no `?install` help in remotes. Maybe just to `?install_github` as the other install functions are now in the package installation family.

vignettes/dependencies.Rmd
```diff
@@ -0,0 +1,91 @@
...
+```yaml
+Remotes: r-lib/testthat
+```
+
+You can also specify a specific hash, tag, or pull request (using the same syntax as `install_github()` if you want a particular commit. Otherwise the latest commit on the master branch is used.
+
+```yaml
+Remotes: r-lib/httr@v0.4,
+  klutometis/roxygen#142,
+  r-lib/testthat@c67018fa4970
```
jimhester: Maybe we should mention the `@*release` syntax here as well? It is not very well known, I am not entirely positive it will even work with remotes actually...

vignettes/dependencies.Rmd
```diff
@@ -0,0 +1,91 @@
...
+Remotes: [type::]<Repository>, [type2::]<Repository2>
+```
+
+The `type` is an optional parameter.  If the type is missing the default is
+to install from GitHub. Additional remote dependencies should be separated by
+commas, just like normal dependencies elsewhere in the `DESCRIPTION` file.
+
+### Github
+
+Because github is the most commonly used unofficial package distribution in R, it's the default:
```
jimhester: Should probably use GitHub's preferred capitalization of GitHub.

vignettes/dependencies.Rmd
```diff
@@ -0,0 +1,91 @@
...
+Remotes: svn::https://github.com/tidyverse/stringr
+
+# URL
+Remotes: url::https://github.com/tidyverse/stringr/archive/master.zip
+
+# Local
+Remotes: local::/pkgs/testthat
+
+# Gitorious
+Remotes: gitorious::r-mpc-package/r-mpc-package
```
jimhester: We don't have Gitorious remotes anymore

gaborcsardi:
Thanks for catching these! Fixing...

vignettes/dependencies.Rmd
gaborcsardi: it used  to work, but yeah, it does not  work now:
```r
❯ install_github("r-lib/R6@*release")
Error in github_resolve_ref.github_release(meta$ref %||% ref, meta, auth_token) :
  No releases found for repo r-lib/R6.
```
https://github.com/r-lib/remotes/issues/200

gaborcsardi:
OK, fixed, merging...

Issue #201, opened 2018-10-12
install_github without submodules


jonasfoe:
I used to instruct for a package which has several large submodules that are for dev purposes / keeping track of dependencies to be installed by remotes.
The submodules used to be ignored.

See: https://github.com/jpahle/CoRC
remotes::install_github("jpahle/CoRC")

Now it seems that this way it will try to pull the submodules.
Even if I do remotes::install_url("https://github.com/jpahle/CoRC/archive/master.zip") it does so:
`Downloading package from url: https://github.com/jpahle/CoRC/archive/master.zip
"C:\tools\git\bin\git.exe" clone --depth 1 --no-hardlinks --recurse-submodules https://github.com/copasi/copasi-dependencies.git C:\Users\*\AppData\Local\Temp\Rtmp8IUAJq\remotes49d04595c3c/CoRC-master/development/copasi-source/copasi-dependencies`

I am currently not sure what it does exactly as I can never get it to finish but I would need to have a way to force the zip download instead of the git pull.

gaborcsardi:
Yes, we now automatically include submodules and AFAIR there is no way to avoid this. We'll need to add an argument, option or env var.  The difficulty is that sometimes you only want to ignore submodules of some packages, but I guess as a start we can just ignore all.

For now you can use the old version of remotes like this:
```
source("https://raw.githubusercontent.com/r-lib/remotes/0a5e4fdc51c1cfb69b071d45bdfa15a96098da0c/install-github.R")$value("jpahle/CoRC")
```

gaborcsardi:
@jimhester I think we can have an argument that suppresses all submodules for a start?

jonasfoe:
Thanks for the workaround. It just doesn't look like one that would inspire confidence in a user. But I'll take it for now.

What I mean is that when I run the aforementioned command I would expect for it to download the zip.
I am using install_url instead of install_github.
Instead it seems to trigger a git pull on the jpahle/CoRC repository.
The output as shown earlier:
`Downloading package from url: https://github.com/jpahle/CoRC/archive/master.zip "C:\tools\git\bin\git.exe" clone --depth 1 --no-hardlinks --recurse-submodules https://github.com/copasi/copasi-dependencies.git C:\Users\*\AppData\Local\Temp\Rtmp8IUAJq\remotes49d04595c3c/CoRC-master/development/copasi-source/copasi-dependencies`

jonasfoe:
[...]

PR #204, opened 2018-10-12
Skip fetching submodules which exist in .Rbuildignore

jimhester:
This code is more complicated than it would otherwise need to be because
the way that `R CMD build` excludes files means that having `^foo$` in
your .Rbuildignore will ignore _all_ files in the foo/ directory.

So to determine if a given path should be ignored we need to match both
the path and it's directories against the patterns.

Fixes https://github.com/r-lib/remotes/issues/201

### DIFF
--- a/R/submodule.R
+++ b/R/submodule.R
@@ -91,6 +91,9 @@ update_submodules <- function(source, quiet) {
   }
   info <- parse_submodules(file)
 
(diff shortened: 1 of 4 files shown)


jimhester:
@gaborcsardi could you take a look at this?

Issue #214, opened 2018-10-18
Error with install_deps and Packrat


ankane:
Hey, I'm running into an error with `install_deps` and Packrat. Instructions to repro:

1. Create a new directory
2. Open an R console and run `packrat::init(); install.packages("remotes")`
3. Create a `DESCRIPTION` file with `Imports: DBI` (or any package)
4. Run `remotes::install_deps()`

Error

```
** testing if installed package can be loaded
Error in file(filename, "r", encoding = encoding) : 
  cannot open the connection
Calls: source -> file
In addition: Warning message:
In file(filename, "r", encoding = encoding) :
  cannot open file 'packrat/init.R': No such file or directory
Execution halted
ERROR: loading failed
* removing ‘/some-path/packrat/lib/x86_64-apple-darwin17.6.0/3.5.1/DBI’
Error in i.p(...) : 
  (converted from warning) installation of package ‘DBI’ had non-zero exit status
```

Platform: Mac
Remotes: 2.0.0
Packrat: 0.4.9-14

This could also be a Packrat issue, but the error doesn't occur with devtools, so figured I'd start here.

cderv:
FWIW, I have the same issue, **when in a packrat project** using https://install-github.me service. ie `source("https://install-github.me/r-dbi/DBI")` or even `source("https://install-github.me/r-lib/remotes")`

Regarding the error, it seems to come from `.Rprofile` of the packrat project. When I delete it, no more issue. It is as if the `.Rprofile` of the project is taken into account when *testing if installed package can be loaded*, but run in a working directory where `source("packrat/init.R")` does not work.

Both have a `safe_install_packages` function that calls `with_rprofile_user` that load the current `.Rprofile` (here the one of packrat project) and add some option. This could be the cause of error. 
in `remote` 📦, it can be deactivated by setting env var `R_REMOTES_NO_ERRORS_FROM_WARNINGS` to true. 

If I do that before `Sys.setenv(R_REMOTES_NO_ERRORS_FROM_WARNINGS="true")` not more error when installing from a source with `remotes` 📦 
```r
# enter packrat project getting the packrat .libPaths
packrat::on()
# By default it is set to false, and we have an error
Sys.setenv(R_REMOTES_NO_ERRORS_FROM_WARNINGS="true")
# issue is when installing from source, from inside a packrat project into the packrat lib
remotes::install_github("r-dbi/DBI")
```

Based on this small debugging lookout, I think the issue is related to this
[...]

PR #215, opened 2018-10-18
Remove reading of local .Rprofile

jimhester:
Including the local .Rprofile causes issues with packrat.

Fixes https://github.com/r-lib/remotes/issues/214

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 
 # devel
 
+* No longer include project specific .Rprofile code in the temporary .Rprofile when
+  `R_REMOTES_NO_ERRORS_FROM_WARNINGS=false` (the default).
+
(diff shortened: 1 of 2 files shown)


gaborcsardi:
So one potential issue with this is that `.Rprofile` might set up a separate library, and then we sould install the packages there. Does packrat do that?

jimhester:
It does, but we are inheriting the libpaths from the parent process, so not including it in the child seems to install in the expected packrat library.

I ran

```r
packrat::on()
remotes::install_github("r-dbi/DBI")
```

With this PR and everything worked without error, and DBI was installed in the packrat library.

gaborcsardi:
Cool, then could you add news and merge please. Thanks much.

PR #216, opened 2018-10-19
Remotes rate limit

jimhester:
So this uses the same bundled PAT on CI systems as devtools and adds more instructions on what to do if you hit the error.

Fixes #210

PR #219, opened 2018-10-23
Fix a bug in github_pat()

yutannihilation:
I guess #216 forgot to delete this early return so the bundled PAT will never be used.

https://github.com/r-lib/remotes/blob/2729d0e6cba9306bbfdfcde9d58133f87de20099/R/github.R#L73

### DIFF
--- a/tests/testthat/test-github.R
+++ b/tests/testthat/test-github.R
@@ -6,6 +6,12 @@ test_that("github_pat", {
 
   expect_equal(github_pat(), "badcafe")
   expect_message(github_pat(quiet = FALSE), "Using github PAT from envvar GITHUB_PAT")
+
+  withr::local_envvar(c(GITHUB_PAT=NA, CI=NA))
+  expect_equal(github_pat(), NULL)
+
+  withr::local_envvar(c(GITHUB_PAT=NA, CI="true"))
+  expect_true(nzchar(github_pat()))
 })
 
 test_that("github_commit", {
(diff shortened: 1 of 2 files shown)


tests/testthat/test-github.R
```diff
@@ -6,6 +6,12 @@ test_that("github_pat", {
 
   expect_equal(github_pat(), "badcafe")
   expect_message(github_pat(quiet = FALSE), "Using github PAT from envvar GITHUB_PAT")
+
+  withr::local_envvar(c(GITHUB_PAT=NA, CI=NA))
```
jimhester:
```suggestion
  withr::with_envvar(c(GITHUB_PAT=NA, CI=NA), {
     expect_equal(github_pat(), NULL)
  })
```

tests/testthat/test-github.R
```diff
@@ -6,6 +6,12 @@ test_that("github_pat", {
...
+  expect_equal(github_pat(), NULL)
+
+  withr::local_envvar(c(GITHUB_PAT=NA, CI="true"))
```
jimhester:
```suggestion
  withr::with_envvar(c(GITHUB_PAT=NA, CI="true"), {
    expect_true(nzchar(github_pat()))
  })
```

yutannihilation:
Thanks for suggestion!

jimhester:
Thanks!

PR #222, opened 2018-10-25
Permit warnings if the envvar is unset and force suggests is false

### LINKED ISSUE #221: Converting warnings into errors has other bad side-effects
e.g. https://github.com/tidyverse/dplyr/pull/3940

jimhester:
A common case when warnings are emitted during installation is when a
Suggested package is not installed. So if
`R_REMOTES_NO_ERRORS_FROM_WARNINGS` is unset but
`_R_CHECK_FORCE_SUGGESTS_` we now turn off converting warnings to
errors.

Fixes #221

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # remotes 2.0.1.9000 - Development
 
+* `install()` now avoids converting warnings to errors if
+  `R_REMOTES_NO_ERRORS_FROM_WARNINGS` is unset and
+  `_R_CHECK_FORCE_SUGGESTS_=false`. This avoids failures due to Suggested
+  packages potentially being missing.
+
 * `install_bitbucket()` now works properly with packages in subdirectories
   (#220)
 
--- a/R/install.R
+++ b/R/install.R
@@ -59,7 +59,7 @@ safe_install_packages <- function(...) {
 
     # Set options(warn = 2) for this process and child processes, so that
     # warnings from `install.packages()` are converted to errors.
-    if (Sys.getenv("R_REMOTES_NO_ERRORS_FROM_WARNINGS") != "true") {
+    if (should_error_for_warnings()) {
       with_options(list(warn = 2),
         with_rprofile_user("options(warn = 2)",
           i.p(...)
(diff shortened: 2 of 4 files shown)


NEWS.md
```diff
@@ -1,5 +1,9 @@
 # remotes 2.0.1.9000 - Development
 
+* `install()` now permits warnings if `R_REMOTES_NO_ERRORS_FROM_WARNINGS` is
```
gaborcsardi: Maybe an  "only"  would be good here. And maybe "now only throws errors for warnings", no?

R/install.R
```diff
@@ -183,3 +183,10 @@ install_deps <- function(pkgdir = ".", dependencies = NA,
     ...
   )
 }
+
+should_error_for_warnings <- function(
+  no_errors = Sys.getenv("R_REMOTES_NO_ERRORS_FROM_WARNINGS", !as.logical(force_suggests)),
+  force_suggests = Sys.getenv("_R_CHECK_FORCE_SUGGESTS_", "true")
+  ) {
+  !as.logical(no_errors)
```
gaborcsardi: Why writing it like this, why not just putting these in function body?

inst/README.markdown
```diff
@@ -197,7 +197,9 @@ will allow successful installation of these packages.
   installation for warning messages. Warnings usually mean installation
   errors, so by default remotes stops for a warning. However, sometimes
   other warnings might happen, that could be ignored by setting this
-  environment variable.
+  environment variable. Warnings are also allowed if
+  `R_REMOTES_NO_ERRORS_FROM_WARNINGS` is unset and `_R_CHECK_FORCE_SUGGESTS_`
+  is `FALSE`.
```
gaborcsardi: I think this is confusing. This paragraph is about suppressing the  warning -> error conversion. So we don't need to modify it. Maybe have a new item for `_R_CHECK_FORCE_SUGGESTS_`?

jimhester:
Thanks for reviewing, I changed the implementation as suggested and tweaked the wording / layout of the news and readme.

gaborcsardi:
Looks good, thanks!

PR #226, opened 2018-10-26
Remove DESCRIPTION metadata from install_version

### LINKED ISSUE #224: `install_version()` should create a CRAN remote, not a URL remote
It uses `install_url()` internally, but I think the metadata should be as if it was a CRAN remote.

jimhester:
install_version calls install_url internally, so these packages were
being inadvertently installed as URL remote packages, which makes them
not update like normal packages.

Now we remove this extraneous metadata, so it is as if they were installed from
CRAN with `install.packages()` originally.

Fixes #224

### DIFF
--- a/R/install-remote.R
+++ b/R/install-remote.R
@@ -77,12 +77,28 @@ install_remotes <- function(remotes, ...) {
 
 # Add metadata
 add_metadata <- function(pkg_path, meta) {
-  path <- file.path(pkg_path, "DESCRIPTION")
-  desc <- read_dcf(path)
 
-  desc <- utils::modifyList(desc, meta)
+  # During installation, the DESCRIPTION file is read and an package.rds file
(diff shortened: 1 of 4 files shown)


R/install-remote.R
```diff
@@ -77,12 +77,29 @@ install_remotes <- function(remotes, ...) {
...
 add_metadata <- function(pkg_path, meta) {
-  path <- file.path(pkg_path, "DESCRIPTION")
-  desc <- read_dcf(path)
 
-  desc <- utils::modifyList(desc, meta)
+  # During installation, the DESCRIPTION file is read and an package.rds file
+  # created with most of the information from the DESCRIPTION file. Functions
+  # that read package metadata may use either the DESCRIPTION file or the
+  # package.rds file, therefore we attempt to modify both of them, and return an
+  # error if neither one exists.
```
gaborcsardi: I don't think we actually report an error here?

PR #227, opened 2018-10-27
Fix for install_deps(upgrade=FALSE) and remote packages

ankane:
Currently on master, `remotes::install_deps(upgrade=FALSE)` does nothing for a `DESCRIPTION` file like:

```
Imports: dbx
Remotes: github::ankane/dbx
```

This fixes it.


ankane:
Looks like one of the Travis test runs needs restarted.

jimhester:
This looks good to me, thanks!

PR #240, opened 2018-11-07
add 'remotes.upgrade' option for upgrade toggle

kevinushey:
This allows users to specify a default upgrade preference (for those of us who might prefer automatically updating dependent packages when installing the development version of a package).


gaborcsardi:
Thanks! Sounds good. However, we already have some environment variables, see https://github.com/r-lib/remotes#environment-variables

Environment variables are better in some ways, e.g. subprocesses inherit them, and you can set them globally for all R sessions.

Do you have a particular reason to have an option instead of an env var?

kevinushey:
Using an R option just seemed the most natural mechanism when I first thought about it, but I'd be entirely open to using environment variables (especially if we want / need to think about subprocesses). Would using e.g. `Sys.getenv("REMOTES_UPGRADE", "ask")` be better?

R/deps.R
```diff
@@ -529,6 +529,14 @@ remote_deps <- function(pkg, ...) {
   res
 }
 
+upgrade_default <- function() {
+
+  env <- Sys.getenv("R_REMOTES_UPGRADE", unset = NA)
```
jimhester: Couldn't this whole function just be

```suggestion
  Sys.getenv("R_REMOTES_UPGRADE", "ask")
```

jimhester:
My main concern with this is you lose the auto-documenting behavior of `c("ask", "always", "never")`, but I don't really see a good way to retain that and allow people to change the default.

gaborcsardi:
We could have `upgrade = c("default", "ask", "always", "never")`, and then `"default"` would do `Sys.getenv("...", if  (interactive()) "ask" else "always")`.

kevinushey:
I've updated the PR with the new `"default"` option -- let me know how it looks.

jimhester:
This LGTM

gaborcsardi:
Thanks!

PR #265, opened 2018-12-03
remotes of remotes not given auth_token

overmar:
When a package listed in Remotes of the description has a package listed in the Remotes, the auth_token wasn't transferring


overmar:
Hold off for a second this stopped working again

PR #266, opened 2018-12-04
Patch 1

overmar:
When install_gitlab has to install packages from the remotes line, it loses its auth_token, this allows the auth_token to persist in the ... and be recalled when necessary

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -53,7 +53,8 @@
 
 package_deps <- function(packages, dependencies = NA,
                          repos = getOption("repos"),
-                         type = getOption("pkgType")) {
+                         type = getOption("pkgType"), 
+                         ...) {
 
   repos <- fix_repositories(repos)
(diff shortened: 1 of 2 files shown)


bestdan:
I think this might solve this issue for github too:
https://github.com/r-lib/remotes/issues/295

Any reason to not make both changes in this PR?

antoine-sachet:
This may have unintended consequences, just like PR #145 introduced #337.

This works when the dependency needs the same host, user, password, auth_token, etc. But what if it does not? Then you've just broken the installation by passing the wrong host. Even if you just pass the credentials, you may pass the wrong credentials: imagine a gitlab package with a dependency on a github package, it will use the wrong auth_token.

You can easily pass credentials to the dependencies using environment variables (and it won't pass the wrong credentials), so there is no need for this in the first place. A better solution would be a useful message if private dependencies are detected/suspected.

PR #268, opened 2018-12-12
Move install script into the package

gaborcsardi:
* Add some minimal tests for it.
* Run it in standalone mode.
* Do not load crancache in standalone mode.
* Test that standalone mode does not load any packages.


gaborcsardi:
@jimhester I think this is ready, if you have some time to take a look.

tests/testthat/test-script.R
```diff
@@ -0,0 +1,61 @@
+
+context("install-github.R script")
+
+test_that("install-github.R script is up to date", {
+  root <- system.file(package = .packageName)
```
jimhester: Maybe we should use `utils::packageName()`? 

> > So, should I use .packageName or utils::packageName() for this?
> 
> You should use utils::packageName.  It is documented; .packageName is not.
> 
> Duncan Murdoch

(https://www.mail-archive.com/r-devel@r-project.org/msg35792.html)

jimhester:
Oh do we want to mention this in the news / README? 

That you can use 

`source(system.file(package = "remotes", "install-github.R")))(args)` to install packages without loading the remotes package?

tests/testthat/test-script.R
gaborcsardi: OK

gaborcsardi:
Thanks!

IDK about the README / NEWS, maybe.

gaborcsardi:
I will not add it to the README, for now, as you don't lose too much by loading the package....

Issue #206, opened 2018-10-12
install_dev should allow @ symbol for refs


wch:
For example, this doesn't work:

```R
install_dev('shiny@v1.2-rc')
#> Error in utils::download.file(url, path, method = download_method(), quiet = quiet,  : 
#>   cannot open URL 'https://cloud.r-project.org/web/packages/shiny@v1.2-rc/DESCRIPTION'
```

But this does:

```R
install_dev('shiny', ref = 'v1.2-rc')
```

mkearney:
DIBS

PR #279, opened 2019-01-19
Allow ref/pull format in install_dev

mkearney:
As requested in #206. Checks and then stores ref/pull format to include in install_repo call.


gaborcsardi:
Thanks! I wonder if we could do this with a single regex. E.g. sg like 
```
re_match(c("foo", "bar", "foobar@ref"), "^(?<pkg>[^@#]+)(?<ref>[@#].*)?$")
```
will split up the input, and then you just add back the ref, like you already do. I would put this in a `dev_split_ref()` function, so that we can test it.

mkearney:
Sounds good! I'm on it!

R/install-dev.R
```diff
@@ -28,15 +28,7 @@ install_dev <- function(package, cran_url = getOption("repos")[["CRAN"]], ...) {
...
-  # if @[ref] is included, save for repo install
-  if (contains_ref(package)) {
-    ref <- sub(".*(?=(@|#))", "", package, perl = TRUE)
-    package <- sub("(#|@).*", "", package)
-  } else {
-    ref <- ""
-  }
-
-  url <- build_url(cran_url, "web", "packages", package, "DESCRIPTION")
+  url <- build_url(cran_url, "web", "packages", dev_split_ref(package)[["pkg"]], "DESCRIPTION")
```
mkearney: I don't know if it's the best way to access "pkg" and "ref" values, but I used double brackets and column names.

tests/testthat/test-utils.R
```diff
@@ -179,3 +179,15 @@ test_that("in_r_build_ignore works", {
...
+  expect_equal(dev_split_ref("DT")[["pkg"]], "DT")
+  expect_equal(dev_split_ref("remotes")[["ref"]], "")
+  expect_equal(dev_split_ref("with.dot")[["pkg"]], "with.dot")
+  expect_equal(dev_split_ref("with2")[["pkg"]], "with2")
+  expect_equal(dev_split_ref("with@v1.2.1")[["ref"]], "@v1.2.1")
+  expect_equal(dev_split_ref("with@v1.0.0.999")[["ref"]], "@v1.0.0.999")
+  expect_equal(dev_split_ref("with@v1.0.0.999")[["pkg"]], "with")
+  expect_equal(dev_split_ref("with#279")[["ref"]], "#279")
+  expect_equal(dev_split_ref("with#1")[["pkg"]], "with")
+})
```
mkearney: These tests all assume input format is `(pkg[(@ref|#pull)]?)`

R/install-dev.R
gaborcsardi: Yes, that's good, but can we call `dev_split_ref` only once? Not a big a deal performance-wise, more of a principle.  Store the result here, and use it later below.

tests/testthat/test-utils.R
gaborcsardi: Yes, these  look great!

jimhester:

Thanks!! You are da bomb!

PR #280, opened 2019-01-19
Improve 404 messages for misspelled repo names/owners or lack of permissions

### LINKED ISSUE #271: Improve error messages when repository not found
Currently it looks like this

``` r
remotes::install_github("r-lib/arst")
#> Error: HTTP error 404.
#>   Not Found
#> 
#>   Rate limit remaining: 4990/5000
#>   Rate limit reset at: 2019-01-04 13:57:37 UTC
#> 
#> 
```

<sup>Created on 2019-01-04 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

Some ideas for improvement

- Only show the rate limit message for 403 errors or when the rate limit is 0 (which should be a 403 error)
- Add a hint when it is a 404 error that it is likely due to a misspelled repo

The relevant code is https://github.com/r-lib/remotes/blob/a0d6cd5d52062fcaedac7dd19b03f9211f578791/R/github.R#L144-L188

### LINKED ISSUE #276: install_github 404 error when Suggest'ed remote is private
I maintain a package to which I've added a `Suggest`ed package that is in a private repository. Even though the default is to not install `Suggest`ed packages, I still receive a 404 error with the following.

```
devtools::install_github("Displayr/flipFormat")
```

I would expect this to work without error, but it would also be nice if the error message could be improved to state the offending package (#271).

tylermorganwall:
I have included code to return the repo name and repo owner in the error message when the status code is 404 (to check for misspellings), as well as remind the user to check if they have the required permissions to access the repo.

Fixes #271.
Fixes #276.

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -165,8 +165,28 @@ github_error <- function(res) {
           "Use `usethis::edit_r_environ()` and add the token as `GITHUB_PAT`."
         }
       )
-  }
+  } else if (identical(as.integer(res$status_code), 404L)) {
+    repo_information = re_match(res$url,"(repos)/(?P<owner>[^/]+)/(?P<repo>[^/]++)/")
+    pat_guidance <- sprintf(
+"Did you spell the repo owner (`%s`) and repo name (`%s`) correctly?
+  - If spelling is correct, check that you have the required permissions to access the repo.",
+        repo_information$owner,
+        repo_information$repo
+    )
 
+  }
+ if(identical(as.integer(res$status_code),404L)) {
+   msg <- sprintf(
+     "HTTP error %s.
+  %s
+
+  %s",
+
+     res$status_code,
+     error_details,
+     pat_guidance
+   )
+ } else {
   msg <- sprintf(
 "HTTP error %s.
   %s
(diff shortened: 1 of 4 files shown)


gaborcsardi:
Thanks! Taking a look.....

R/github.R
```diff
@@ -165,8 +165,28 @@ github_error <- function(res) {
           "Use `usethis::edit_r_environ()` and add the token as `GITHUB_PAT`."
         }
       )
+  } else if (identical(as.integer(res$status_code), 404L)) {
+    repo_information = re_match(res$url,"(repos)/(?P<owner>[^/]+)/(?P<repo>[^/]++)/")
```
gaborcsardi: Can you pls use `<-` and space after `,`?

R/github.R
```diff
@@ -165,8 +165,28 @@ github_error <- function(res) {
...
+    pat_guidance <- sprintf(
```
gaborcsardi: Can you rename `pat_guidence` to just `guidence`. It is now not always about the PAT.

R/github.R
tylermorganwall: Done.

R/github.R
tylermorganwall: Done.

gaborcsardi:
Looks good, thanks much!

QuLogic:
A few files were inadvertently made executable here.

gaborcsardi:
@QuLogic can you pls open an issue or PR? Thanks!

QuLogic:
OK #341

PR #285, opened 2019-01-23
Add host argument to github_resolve_ref()

### LINKED ISSUE #284: github_resolve_ref() should take `host`

yutannihilation:
Fix #284 

Currently, `github_resolve_ref()` doesn't pass `host` to `github_GET()`, so if the ref contains a branch or a PR name, it fails for GitHub Enterprise.

This PR adds `host` argument to `github_resolve-ref()` and pass `host` to it.

### DIFF
--- a/R/install-github.R
+++ b/R/install-github.R
@@ -77,7 +77,7 @@ github_remote <- function(repo, ref = "master", subdir = NULL,
                        host = "api.github.com", ...) {
 
   meta <- parse_git_repo(repo)
-  meta <- github_resolve_ref(meta$ref %||% ref, meta, auth_token)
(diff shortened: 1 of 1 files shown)


yutannihilation:
Note that, though I believe this is the right change to make, I cannot use `devtools::install_github()` without `dependencies = FALSE` for a GitHub Enterprise that requires PATs. We can set `GITHUB_PAT` for only one of the GitHub Enterprise or github.com, which means the other one fails with 401 error (bad credentials)...

I almost gave up. No idea what to do to make remotes do the right thing.

jimhester:
I don't think we can support multiple PATs without serious refactoring of the internals. I guess the assumption is if you are using github enterprise all of the github remotes should live in your enterprise instance.

PR #300, opened 2019-02-20
Retain attributes when combining cran_deps + remote_deps

jennybc:
`cran_deps` bears the `type` attribute

This fixes the problem I described in slack.

The order of arguments in `rbind()` affects the attributes of the resulting data frame.

``` r
a <- data.frame(a = 1)
attr(a, "yo") <- "hi"

b <- data.frame(a = 2)

attr(rbind(a, b), "yo")
#> [1] "hi"
attr(rbind(b, a), "yo")
#> NULL
```

<sup>Created on 2019-02-19 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

I don't know why other people aren't seeing this, so perhaps there is more going on? But this fixes things for me.

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -150,8 +150,7 @@ combine_deps <- function(cran_deps, remote_deps) {
 
   # If there are remote deps remove the equivalent CRAN deps
   cran_deps <- cran_deps[!(cran_deps$package %in% remote_deps$package), ]
-
-  rbind(remote_deps, cran_deps)
+  rbind(cran_deps, remote_deps)
 }
 
 ## -2 = not installed, but available on CRAN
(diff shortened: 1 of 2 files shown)


jennybc:
This PR passed tests (with 1 skipped test) everywhere but R-devel on travis, so I think that difficulty is orthogonal to this.

jennybc:
Maybe getting closer to the root cause ...

I think this row indexing is where the `type` attribute can potentially be lost from either the cran or gh deps.

https://github.com/r-lib/remotes/blob/62123450b3b6c1c7c485e88a3b204dbc21166100/R/deps.R#L148-L152

``` r
df <- data.frame(a = 1:3)
attr(df, "yo") <- "hi"
attr(df, "yo")
#> [1] "hi"
attr(df[1, ], "yo")
#> NULL
```

<sup>Created on 2019-02-20 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

jennybc:
Closing in favour of https://github.com/r-lib/remotes/issues/304

Issue #259, opened 2018-11-21
Unable to `install_gitlab` repos within subgroups 


timtrice:
If we want to use `install_gitlab` we supply a `repo` parameter which is in <code>username/repo[/subdir][@ref]</code> format.  This string is split by [parse_repo_spec](https://github.com/r-lib/remotes/blob/fb6682b6091412e75e313345ab54e96e82ed58b7/R/parse-git.R#L40) so we have a `username` variable (e.g., "r-lib") and `repo` variable (e.g., "remotes"). 

If my repo is within a  GitLab group, then I can simply replace the `username` portion of the `repo` parameter with the name of the group.

However, if I have a subgroup holding the repo, I'm stuck. There is no `params` variable for subgroups.

```r
#' Works
devtools::install_gitlab("timtrice/falsy")
```

```r
#' Works
devtools::install_gitlab("timtrice_tests/falsy")
```

```r
#' Fails
devtools::install_gitlab("timtrice_tests/tests/falsy")
```

```
 Error in utils::download.file(url, path, method = download_method(), quiet = quiet,  : 
  cannot open URL 'https://gitlab.com/timtrice_tests/tests/raw/master/falsy/DESCRIPTION?private_token=ABCDEFGHIJKLMNOP' 
```

There is no way to get the subgroup past the `parse_repo_spec` function. So, further functions that stick the URL back together do it incorrectly. The URL in the error above should be, 

<code>https://gitlab.com/timtrice_tests/tests/raw/master/falsy/DESCRIPTION?private_token=ABCDEFGHIJKLMNOP</code>

For example, in `remote_package_name`, the src is built (using `build_url`) as that in the error message. But, if I change src to the last example above, while debugging, it is successful. 

Another option is, instead of using the group/subgroup/repo combination, perhaps we can pass the Project ID? This doesn't seem to work in `remote_pacage_name` but it does in `gitlab_commit`.

jimhester:
I was not aware of this subgroup concept, looking at this example I am not really sure how it could be supported while also retaining support for subdirectories

rozwad:
Please see #308 – quick and dirty workaround.

ldecicco-USGS:
Any chance we can re-start the conversation on installing from sub groups? We're considering moving our github/USGS-R packages to a sub group. Here's an example:

The subgroup:
https://code.usgs.gov/water/stats

The repo:
https://code.usgs.gov/water/stats/MGBT

For the record, this works:
```r
[...]

PR #308, opened 2019-03-10
Support for GitLab subgroups

rozwad:
A quick and dirty prototype workaround to address issue #259.

As the `install_gitlab` seems to not support (yet) subdirectories, the workaround is based on the `subdir` variable. However in case a repo is inside a subgroup there is a mismatch – `subdir` corresponds to a repository, and `repo` to a subgroup (result of the general parsing rules in [parse_repo_spec](https://github.com/r-lib/remotes/blob/fb6682b6091412e75e313345ab54e96e82ed58b7/R/parse-git.R#L40)). How about fixing that mismatch with a new dedicated `parse_repo_gitlab` function?

### DIFF
--- a/R/install-gitlab.R
+++ b/R/install-gitlab.R
@@ -70,12 +70,12 @@ gitlab_remote <- function(repo,
 remote_download.gitlab_remote <- function(x, quiet = FALSE) {
   dest <- tempfile(fileext = paste0(".tar.gz"))
 
-  src_root <- build_url(x$host, x$username, x$repo)
+  src_root <- build_url(x$host, x$username, x$repo, x$subdir)
   src <- paste0(src_root, "/repository/archive.tar.gz?ref=", utils::URLencode(x$ref, reserved = TRUE))
 
   if (!quiet) {
-    message("Downloading GitLab repo ", x$username, "/", x$repo, "@", x$ref,
-            "\nfrom URL ", src)
+    message("Downloading GitLab repo ", gsub("/$", "", x$username, "/", x$repo, "/", x$subdir),
+            "@", x$ref, "\nfrom URL ", src)
   }
 
   download(dest, src, auth_token = x$auth_token, auth_phrase = "private_token=")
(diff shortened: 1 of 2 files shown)


R/install-gitlab.R
```diff
@@ -70,12 +70,12 @@ gitlab_remote <- function(repo,
...
   dest <- tempfile(fileext = paste0(".tar.gz"))
 
-  src_root <- build_url(x$host, x$username, x$repo)
+  src_root <- build_url(x$host, x$username, x$repo, x$subdir)
   src <- paste0(src_root, "/repository/archive.tar.gz?ref=", utils::URLencode(x$ref, reserved = TRUE))
 
   if (!quiet) {
-    message("Downloading GitLab repo ", x$username, "/", x$repo, "@", x$ref,
-            "\nfrom URL ", src)
+    message("Downloading GitLab repo ", gsub("/$", "", x$username, "/", x$repo, "/", x$subdir),
```
jimhester: Are you actually getting a trailing slash in the subdirectory or just being cautious? In general I would consider specs with a trailing slash to be invalid input, and actually it seems `parse_git_repo()` strips them anyway, so you can remove the gsub here and below I think.

```r
> parse_git_repo("foo/bar/baz/")
$username
[1] "foo"

$repo
[1] "bar"

$subdir
[1] "baz"
```

jimhester: Oh now that I look at this again this handles the case when subdir is empty. I think a better approach is to use `paste0(c(x$username, x$repo, x$subdir), collapse = "/")`

robertdj:
I am also missing this functionality of installing packages from a path with three levels (not sure if it is a subgroup or subdirectory).

Having tried out this branch I still run into problems.

As far as I can see `install_gitlab` makes three calls to get the repository in `path/to/repo`:

- One to get `DESCRIPTION`: https://gitlab.foo.com/path/to/repo/raw/master/DESCRIPTION?private_token=token
- One to get some metadata about the repo using a URL encoded `path/to/repo`: https://gitlab.foo.com/api/v4/projects/path%2Fto%2Frepo/repository/commits/master?private_token=token
- One to get the repository as a `tar.gz` file: https://gitlab.foo.com/path/to/repo/repository/archive.tar.gz?ref=master?private_token=token

The first two works for me, but the last one fails.
According to [GitLab's documentation](https://docs.gitlab.com/ee/api/repositories.html#get-file-archive) the last call should also be different -- using the same form as the second call:
https://gitlab.foo.com/api/v4/projects/path%2Fto%2Frepo/repository/archive.tar.gz?ref=master&private_token=<token>

With this change it works for me. 
@rozwad : Does this also work for you?

kevinushey:
IMHO we should avoid overloading the meaning of `subdir` (since a subgroup and folder within a repository are different things) and instead consider how to handle sub-directories and sub-groups.

I think it would be better to augment the parser so that there's a way to parse sub-directories and sub-groups separately (either globally or just a GitLab-only parser).

PR #314, opened 2019-03-14
README tweaks

maelle:
- typo fixes
- link to the rendered version of the mentioned vignette
- and a question: I read "BioConductor packages are automatically detected and their dependencies are installed from BioConductor. The BiocInstaller package, which is needed to install them, is also automatically installed temporarily." but it doesn't seem the `BiocInstaller` package is installed at all?


jimhester:
Thanks!

Issue #313, opened 2019-03-14
install_bioc() without git?


maelle:
Should `install_bioc()` (or a second function) also offer the possibility to install from BioConductor repository via `install.packages()` (like what `BiocManager` does) rather than with `git`? For the example below, installation actually fails with `install_bioc()`(not `remotes`' fault, when running this outside of `reprex` I saw it was a compilation error, in the two cases the source is probably different)

``` r
repos <- structure(c("https://cloud.r-project.org", "https://bioconductor.org/packages/3.6/bioc", 
                     "https://bioconductor.org/packages/3.6/data/annotation", "https://bioconductor.org/packages/3.6/data/experiment"
), .Names = c("CRAN", "BioCsoft", "BioCann", "BioCexp"))

install.packages("IRanges", repos = repos)
#> Installing package into '/home/maelle/R/x86_64-pc-linux-gnu-library/3.4'
#> (as 'lib' is unspecified)
remove.packages("IRanges")
#> Removing package from '/home/maelle/R/x86_64-pc-linux-gnu-library/3.4'
#> (as 'lib' is unspecified)
remotes::install_bioc("IRanges")
#> Downloading Bioconductor repo https://git.bioconductor.org/packages/IRanges
#> Warning in is.na(remote_deps$package): is.na() applied to non-(list or
#> vector) of type 'NULL'
#> 
#>   
   checking for file ‘/tmp/RtmpLkbH4n/file41571345ef03/DESCRIPTION’ ...
  
✔  checking for file ‘/tmp/RtmpLkbH4n/file41571345ef03/DESCRIPTION’ (1000ms)
#> 
  
─  preparing ‘IRanges’:
#> 
  
   checking DESCRIPTION meta-information ...
  
✔  checking DESCRIPTION meta-information
#> 
  
─  cleaning src
#> 
  
─  checking for LF line-endings in source and make files and shell scripts (375ms)
#> 
  
─  checking for empty or unneeded directories
#> 
  
─  building ‘IRanges_2.16.0.tar.gz’
#> 
  
   
#> 
#> Installing package into '/home/maelle/R/x86_64-pc-linux-gnu-library/3.4'
#> (as 'lib' is unspecified)
#> Error in i.p(...): (converted from warning) installation of package '/tmp/RtmpLkbH4n/file415720227a5d/IRanges_2.16.0.tar.gz' had non-zero exit status
```

[...]

Issue #275, opened 2019-01-15
Skipping packages from Bioconductor


llrs:
The package sessioninfo is able to find which packages are from Bioconductor, but remotes seems to consider everything outside github as ahead of CRAN:

https://github.com/r-lib/remotes/blob/bd970d723facc6ede2ff0f54ce4ae917526a37f8/R/deps.R#L279

So if users of Bioconductor update the packages using remotes (or usually devtools) they get an incorrect message. It would be better if it skips those that are in Bioconductor and redirect them to the BiocManager package with a sensible message.

llrs:
Related to this, there have been a discussion recently about installing Bioconductor packages from source and binaries that might be relevant (not sure if here or an independent issue)

See this thread: https://stat.ethz.ch/pipermail/bioc-devel/2020-April/016689.html

PR #315, opened 2019-03-14
Bioc version

maelle:
# Proposed changes

- Not in this PR, having an install_bioc() function that'd be more similar to the current install_cran() function, installing without git. Cf https://github.com/r-lib/remotes/issues/313

- Using a logic similar to BiocManager in bioc.R but with also the use of an environment variable that one could use to force the use of Bioconductor devel version.

- At the moment tests don't pass because of several factors
    - for some R versions it seems I'm not deducing the same Bioconductor version (but my code is in agreement with BiocManager)
    - on my computer there are default repositories for Bioconductor 3.6 and with the current code it interferes with the right Bioconductor version.

# Bioconductor support in remotes

- bioc.R and install-bioc.R are independent: install-bioc.R is only for the install_bioc() function, and doesn't use the other Bioconductor-related script, bioc.R.

- install_bioc() uses *git*. You can specify a mirror via the "BioC_git" option, but you cannot specify a Bioconductor version.

- As dependencies, Bioconductor packages are installed from the Bioconductor repository corresponding to the Bioconductor version determined via some logic. They are viewed as CRAN packages which apparently can lead to some issues cf https://github.com/r-lib/remotes/issues/275

# Bioconductor and R in general

- The installation of Bioconductor packages used to happen via a script sourced from an URL, which has been replaced with a proper CRAN package, BiocManager, available for R >= 3.5.0.

- Good read about the non-security of the old approach: https://www.jumpingrivers.com/blog/security-r-hacking-bioconductor/

- In R itself there's a legacy environment variable R_BIOC_VERSION https://github.com/wch/r-source/search?q=R_BIOC_VERSION&unscoped_q=R_BIOC_VERSION that actually isn't used to install Bioconductor packages for R >= 3.5.0.

- The mapping between R and Bioconductor version is stored in a YAML config file that can be parsed using code from BiocManager. 

- In particular, for some R versions there can be two valid Bioconductor versions at a time, one being the release version and the other being the devel version. A developer can prefer packages to be installed from any of those, so there needs to be a way to choose.
    - The BiocInstaller package had `BiocInstaller::isDevel()` that downloads the dev version of BiocInstaller ensuring all subsequent installations will be from Bioconductor devel version.
    - By default BiocManager uses the Bioconductor release version for the R version, if not available the devel version, if not available an out-of-date version (it's Bioconductor that's out-of-date, but still compatible with the R version), if not avaiable a future version.
    - `BiocManager::install()` has a version argument to specify the Bioconductor version. To switch to "devel" you'd write `BiocManager::install(version = "devel")`.

### DIFF
--- a/R/bioc.R
+++ b/R/bioc.R
@@ -148,6 +148,17 @@ bioc_repos <- function(bioc_ver, r_ver) {
 bioc_install_repos <- function(r_ver = getRversion(),
                                bioc_ver = Sys.getenv("BIOCONDUCTOR_VERSION",
                                                      bioc_version(r_ver))) {
+
+  bioc_ver <- package_version(bioc_ver)
+  # check version compatibility
+  map <- bioc_version_map(r_ver)
+
+  if (!bioc_ver %in% map$Bioc[map$R == r_ver[, 1:2],]){
+    stop("Incompatible R and Bioconductor version, check BIOCONDUCTOR_VERSION",
+         call. = FALSE)
+  }
+
+  # get repos
   repos <- bioc_repos(bioc_ver, r_ver)
 
   repos
--- a/R/bioc.R
+++ b/R/bioc.R
@@ -1,20 +1,104 @@
-## This is mostly from https://bioconductor.org/biocLite.R
+## This is mostly from https://github.com/Bioconductor/BiocManager/blob/ba18f67fb886048b991c0f04a87894cf8c35b076/R/version.R
 
-bioc_version <- function() {
-  bver <- get(
-    ".BioC_version_associated_with_R_version",
-    envir = asNamespace("tools"),
-    inherits = FALSE
-  )
+bioc_version <- function(r_ver) {
+  map <- bioc_version_map(r_ver)
 
-  if (is.function(bver)) {
-    bver()
-  } else {
-    bver
+   if (r_ver[, 1:2] < min(map$R)){
+      stop("Unsupported R version for Bioconductor", call. = FALSE)
+    } else {
+    map <- map[map$R == r_ver[, 1:2],]
+    if ("release" %in% map$BiocStatus)
+      idx <- map$BiocStatus == "release"
+    else if ("devel" %in% map$BiocStatus)
+      idx <- map$BiocStatus == "devel"
+    else if ("out-of-date" %in% map$BiocStatus)
+      idx <- map$BiocStatus == "out-of-date"
+    else
+      idx <- map$BiocStatus == "future"
+
+    return(tail(map$Bioc[idx], 1))
   }
+
 }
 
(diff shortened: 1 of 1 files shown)


maelle:
For info, version map from `BiocManager`

``` r
BiocManager:::.version_map()
#>    Bioc    R  BiocStatus
#> 1   1.6  2.1 out-of-date
#> 2   1.7  2.2 out-of-date
#> 3   1.8  2.3 out-of-date
#> 4   1.9  2.4 out-of-date
#> 5   2.0  2.5 out-of-date
#> 6   2.1  2.6 out-of-date
#> 7   2.2  2.7 out-of-date
#> 8   2.3  2.8 out-of-date
#> 9   2.4  2.9 out-of-date
#> 10  2.5 2.10 out-of-date
#> 11  2.6 2.11 out-of-date
#> 12  2.7 2.12 out-of-date
#> 13  2.8 2.13 out-of-date
#> 14  2.9 2.14 out-of-date
#> 15 2.10 2.15 out-of-date
#> 16 2.11 2.15 out-of-date
#> 17 2.12  3.0 out-of-date
#> 18 2.13  3.0 out-of-date
#> 19 2.14  3.1 out-of-date
#> 20  3.0  3.1 out-of-date
#> 21  3.1  3.2 out-of-date
#> 22  3.2  3.2 out-of-date
#> 23  3.3  3.3 out-of-date
#> 24  3.4  3.3 out-of-date
#> 25  3.5  3.4 out-of-date
#> 26  3.6  3.4 out-of-date
#> 27  3.7  3.5 out-of-date
#> 28  3.8  3.5     release
#> 29  3.9  3.6       devel
#> 30  3.9  3.7      future
```

<sup>Created on 2019-03-14 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

At the moment in `remotes` tests there's 

```r
 expect_equal(
    bioc_install_repos("3.1.0"),
    bioc_repos("2.14")
  )
```
but the table above would choose Bioconductor 3.0, not 2.14, because it's the most recent Bioconductor version compatible with R3.1.0.

maelle:
And the environment variable should maybe be related to the use of devel, not to a version number.

jimhester:
`install_bioc()` is supposed to install from git, it is equivalent to `install_github()` not `install_cran()` and is intended mainly for Bioconductor package developers for use in the `Remotes:` field during development of multiple connected packages.

If you want to install bioconductor packages from the main repository you should just use either `install_cran(repos = BiocManager::repositories())` or use `BiocManager::install()` directly.

LiNk-NY:
@maelle 
Why not import `BiocManager`? This would avoid errors like: 

r-lib/revdepcheck#173

PR #329, opened 2019-04-04
updating so install robust to empty .gitmodules file

jsilve24:
I ran into a small problem when trying to install a package with an empty .gitmodules file. I isolated the problem to a single line in the update_submodules function.

I have corrected the problem by just adding a condition that if the submodules file is empty, then the function should exit in exactly the same way as if the file had not been found. This makes the function robust to empty .gitmodules files. 

Hope this helps. 

Justin

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 
 # remotes dev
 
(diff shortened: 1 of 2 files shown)


jimhester:
Thanks!

PR #340, opened 2019-04-09
Do not rely on the type or repos attribute

### LINKED ISSUE #304: Make "type" less vulnerable to loss
The conclusion of the conversation represented by https://github.com/r-lib/remotes/pull/300 (there and in slack) is that storing `type` as an attribute of these data frames is too brittle. The row subsetting and binding those data frames go through makes it hard to retain the attributes correctly. I'm closing that PR in deference to some larger rethinking of how `type` gets passed around.

Probably also what's going on in #291. That also seems like a case where the specific combination of packages to be installed from various sources was unlucky.

jimhester:
It is too fragile and gets removed when rbinding

Fixes #304


stelsemeyer:
Thanks for fixing this! 😊 👍

PR #345, opened 2019-04-10
Fix/deps host

### LINKED ISSUE #337: wrong host passed to dependencies
The `host` passed at top level is passed to the dependencies. This causes the install to fail when the dependency host should be different.

Consider:

* Package A on bitbucket. Its `DESCRIPTION` has a `Remotes` field with `github::antoine-sachet/packageB`

* Package B on github.

Installing package A with `install_bitbucket("bbuser/packageA")` fails! 

Indeed, with the `host` in `...` being passed to the dependencies, I end up with a `github_remote` with host = "api.bitbucket.org/2.0". Incidentally, this results in #322. 

This issue was introduced by #145 ... It's a bad idea to pass the host to dependencies. I thought the correct way to specify the dependencies' hosts was through the "Remotes" field of the DESCRIPTION?

This is also to some extent a security concern. My credentials are passed to a dependency's host if the corresponding `XXX_remote()` matches the arguments of `install_YYY()`.

antoine-sachet:
This fixes #337 which currently prevents the installation of a package with dependencies hosted on a different service (e.g. a bitbucket package with github dependencies). It is effectively a revert of #145.

Important note: this will break the installation of private dependencies for people passing `auth_token` to `install_github` and expecting it to be passed to the dependencies as well.  Which is why I do not expect this PR to be merged (but see next steps below)

Even if we just passed credentials (and not the host) to the dependencies, the install would still break when there is a collision in argument name. Typically, the wrong auth_token would be passed to the github dependency of a gitlab package.

There is however a very simple way to pass credentials to dependencies via environment variables.

Next steps: There is a more elaborate fix. Pass credentials (not host) to the dependency IF AND ONLY IF the dependency host matches the package host. This would not break existing installs so is really a better way to go.


jimhester:
Thank you for the diagnosis and fix. I actually think we should just merge this as-is for now, as you said the environment variables give people a more robust fallback and the current behavior is just broken with remotes from different sources than the top level `install_*()` function.

PR #344, opened 2019-04-10
fix/bitbucket host sha

antoine-sachet:
Fix #342 and #343

PR #348, opened 2019-04-10
Use default bitbucket credentials for bitbucket dependencies

### LINKED ISSUE #347: Bitbucket dependencies should use credentials from environment variables
When creating a `bitbucket_remote`, the default user and password should be fetched from the environment variables. This is consistent with the current behaviour of `github_remote` or `gitlab_remote` which respectively use `auth_token = github_pat()` and `auth_token = gitlab_pat()`.

This is easily fixed in `install-bitbucket.R` by changing:

```
bitbucket_remote <- function(repo, ref = "master", subdir = NULL,
                             auth_user = NULL, password = NULL,
                             sha = NULL, host = "api.bitbucket.org/2.0", ...)
```

to: 

```
bitbucket_remote <- function(repo, ref = "master", subdir = NULL,
                             auth_user = bitbucket_user(), password = bitbucket_password(),
                             sha = NULL, host = "api.bitbucket.org/2.0", ...)
```

In fact it should really have been part of PR #344 but I only noticed it now. Until this is fixed, there is no way to install a private bitbucket repository since credentials are no longer passed to dependencies.

antoine-sachet:
Fix #347. Complement PR #344.

Until this is merged, there is no way to install a private bitbucket dependency since credentials are no longer passed to dependencies. We need the dependencies to use the environment variables.


meztez:
Would you like some help with this pull request, we are also experiencing the same behavior and I came to the same conclusion as you, but I don't want to create a separate pull request for the same thing. Thanks

jimhester:

Thanks!! You are da bomb!

PR #351, opened 2019-04-13
Remove duplicated example

arnima-github:
Thanks for developing and maintaining 'remotes'.

Well impressed that it requires no additional packages!

Just noticed a duplicated example that can be smoothed out.


gaborcsardi:
Thanks!

Issue #355, opened 2019-04-24
GitLab personal access tokens need api scope


aornugent:
Hia,

`install_gitlab` attempts to use a url encoded path to the repo (see: https://github.com/r-lib/remotes/issues/327)  and failing:

```{r}
remotes::install_gitlab("jimhester/covr")
Error in utils::download.file(url, path, method = download_method(), quiet = quiet,  : 
  cannot open URL 'https://gitlab.com/api/v4/projects/jimhester%2Fcovr/repository/commits/master
```

Navigating directly to [https://gitlab.com/jimhester%2Fcovr](https://gitlab.com/jimhester%2Fcovr) doesn't work either. [https://gitlab.com/jimhester/covr](https://gitlab.com/jimhester/covr) does.

aornugent:
After some snooping it appears the issue isn't URL encoding, rather the personal access token here: https://github.com/r-lib/remotes/blob/1f657ec067088add76adedfcc9a0ea2a45aac9e9/R/install-gitlab.R#L138

```
# Works
gitlab_commit("jimhester", "covr", pat = NULL)
[1] "257a963aaf9b5bdbafc2cec85f14b95ed86a8561"

# Doesn't work
gitlab_commit("jimhester", "covr", pat = gitlab_pat())
Error in utils::download.file(url, path, method = download_method(), quiet = quiet,  : 
  cannot open URL 'https://gitlab.com/api/v4/projects/jimhester%2Fcovr/repository/commits/master?private_token=[removed]' 
```

The [GitLab docs](https://docs.gitlab.com/ee/api/#personal-access-tokens) suggest the API call is well formed, so it's unclear what is going wrong. 
Happy to create a pull request if someone can point me in the right direction.

aornugent:
My bad. I had the wrong scope set on my access token.

For future reference: [Limiting scopes of a personal access token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html#limiting-scopes-of-a-personal-access-token)

trafficonese:
I am still getting the same error from your first post although my personal access tokes has api scope.

This is my (anonymized) command:
`install_gitlab(repo = "user/repo", auth_token = "********", host = "gitlab.mycompany.eu")`

And I am still getting this error:

> Error in utils::download.file(url, path, method = download_method(), quiet = quiet,  : 
>   cannot open URL 'https://gitlab.mycompany.eu/api/v4/projects/user%2Frepo/repository/commits/master?private_token=********

[...]

PR #359, opened 2019-04-26
Fix remote_download url to use GitLab API

aornugent:
`install_gitlab` uses the GitLab API to get commit IDs but not for downloading the repository. This causes it to fail on private repositories, see [this issue on the Rstudio Discourse](https://community.rstudio.com/t/how-to-install-gitlab-from-a-private-repository/26801).

### DIFF
--- a/R/install-gitlab.R
+++ b/R/install-gitlab.R
@@ -70,7 +70,7 @@ gitlab_remote <- function(repo,
 remote_download.gitlab_remote <- function(x, quiet = FALSE) {
   dest <- tempfile(fileext = paste0(".tar.gz"))
 
-  src_root <- build_url(x$host, x$username, x$repo)
+  src_root <- build_url(x$host, "api", "v4", "projects", utils::URLencode(paste0(x$username, "/", x$repo), reserved = TRUE))
   src <- paste0(src_root, "/repository/archive.tar.gz?ref=", utils::URLencode(x$ref, reserved = TRUE))
 
   if (!quiet) {
--- a/R/install-gitlab.R
+++ b/R/install-gitlab.R
@@ -71,7 +71,7 @@ remote_download.gitlab_remote <- function(x, quiet = FALSE) {
   dest <- tempfile(fileext = paste0(".tar.gz"))
 
   src_root <- build_url(x$host, "api", "v4", "projects", utils::URLencode(paste0(x$username, "/", x$repo), reserved = TRUE))
-  src <- paste0(src_root, "/repository/archive.tar.gz?ref=", utils::URLencode(x$ref, reserved = TRUE))
+  src <- paste0(src_root, "/repository/archive.tar.gz?sha=", utils::URLencode(x$ref, reserved = TRUE))
 
   if (!quiet) {
     message("Downloading GitLab repo ", x$username, "/", x$repo, "@", x$ref,
--- a/R/install-gitlab.R
+++ b/R/install-gitlab.R
@@ -106,16 +106,27 @@ remote_metadata.gitlab_remote <- function(x, bundle = NULL, source = NULL, sha =
 remote_package_name.gitlab_remote <- function(remote, ...) {
 
   tmp <- tempfile()
-  src <- build_url(
-    remote$host, remote$username, remote$repo, "raw",
-    remote$ref, remote$subdir, "DESCRIPTION")
+
+  src_root <- build_url(
+    remote$host, "api", "v4", "projects",
+    utils::URLencode(paste0(remote$username, "/", remote$repo),
+                     reserved = TRUE),
+    "repository")
+
+  src <- paste0(
+    src_root, "/files/",
+    ifelse(
+      is.null(remote$subdir),
+      "DESCRIPTION",
+      utils::URLencode(paste0(remote$subdir, "/DESCRIPTION"), reserved = TRUE)),
+    "/raw?ref=", remote$ref, "&private_token=", remote$auth_token)
 
   dest <- tempfile()
   res <- download(dest, src, auth_token = remote$auth_token, auth_phrase = "private_token=")
 
   tryCatch(
     read_dcf(dest)$Package,
-    error = function(e) NA_character_)
+    error = function(e) remote$repo)
 }
 
 #' @export
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,9 @@
 
 * Fix parsing of Additional_Repositories which have a leading newline
   (@tmelliott, #251).
+  
+* Fix API call for private repositories in `install_gitlab` 
+  (@aornugent, [https://community.rstudio.com/t/how-to-install-gitlab-from-a-private-repository/26801](https://community.rstudio.com/t/how-to-install-gitlab-from-a-private-repository/26801)
 
 # remotes 2.0.3
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,7 +20,7 @@
   (@tmelliott, #251).
   
 * Fix API call for private repositories in `install_gitlab` 
-  (@aornugent, [https://community.rstudio.com/t/how-to-install-gitlab-from-a-private-repository/26801](https://community.rstudio.com/t/how-to-install-gitlab-from-a-private-repository/26801)
+  (@aornugent, [https://community.rstudio.com/t/how-to-install-gitlab-from-a-private-repository/26801](https://community.rstudio.com/t/how-to-install-gitlab-from-a-private-repository/26801))
 
 # remotes 2.0.3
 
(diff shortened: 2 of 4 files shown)


dpprdan:
Nice! And such an easy fix.
However, the `ref` parameter is ignored at the moment, i.e.

```r 
install_gitlab("user/repo@non-default-branch")
```

will download and install from the default branch (usually `master`), so the `@non-default-branch` will be ignored. 

According to the [Repository API docs](https://docs.gitlab.com/ce/api/repositories.html#get-file-archive), the branch (or tag or SHA) has to be specified with the `sha` parameter. 

Long story short, [line 74](https://github.com/r-lib/remotes/blob/5415abea005ece6aae93a20d38e8161347b60999/R/install-gitlab.R#L74) 

```r
src <- paste0(src_root, "/repository/archive.tar.gz?ref=", utils::URLencode(x$ref, reserved = TRUE))
```

probably ought to be

```r
src <- paste0(src_root, "/repository/archive.tar.gz?sha=", utils::URLencode(x$ref, reserved = TRUE))
```

Installing from a subdir like e.g. 

```r
install_gitlab("dpprdan/rpostgresql/RPostgreSQL")
```

works BTW, even though it apparently downloads the whole repo and not only the subdir.

aornugent:
Thanks for the feedback. If the API is after a commit ID is there any reason to give it `ref` over `sha`? It gets fetched anyway, but I can't tell if adding it to the `remote` has any obvious consequences, but I am only across the GitLab section of this package.

dpprdan:
Sorry, I don't follow. 

> If the API is after a commit ID is there any reason to give it `ref` over `sha`?

Again, not sure if you meant this, but FWIW, I find Gitlab's API parameters in this particular context a bit confusing. For the [`/projects/:id/repository/archive`](https://docs.gitlab.com/ce/api/repositories.html#get-file-archive) endpoint, the `sha` parameter "[a] tag, branch reference, or [commit] SHA can be used." For the "[list repository tree](https://docs.gitlab.com/ce/api/repositories.html#list-repository-tree)" (the `/projects/:id/repository/tree` endpoint) the `ref` parameter specifies the same things (it only mentions tag or branch name in the docs, but a commit SHA works as well). 

For the `...archive` endpoint `ref` is not defined and effectively ignored, so "This defaults to the tip of the default branch if not specified."  

> It gets fetched anyway and I can't see any obvious consequences of adding it to the `remote`, but I am only across the GitLab section of this package.

Well, if you want to install a branch other than the default (usually `master`) then you would have to specify this as `@other_branch` in the `remote` argument of `install_gitlab`. `install_github` also has a [`ref` argument](https://remotes.r-lib.org/reference/install_github.html#arguments), where you can specify the branch, tag or commit as well (but apparently also in `repo`).

R/install-gitlab.R
```diff
@@ -71,7 +71,7 @@ remote_download.gitlab_remote <- function(x, quiet = FALSE) {
   dest <- tempfile(fileext = paste0(".tar.gz"))
 
   src_root <- build_url(x$host, "api", "v4", "projects", utils::URLencode(paste0(x$username, "/", x$repo), reserved = TRUE))
-  src <- paste0(src_root, "/repository/archive.tar.gz?ref=", utils::URLencode(x$ref, reserved = TRUE))
```
dpprdan: I think there was a misunderstanding. IMO you only need to change the `?ref` to `?sha` and leave everything else as is (or rather revert all other changes you made in this commit).

aornugent:
Maybe I over complicated things sorry. I understood what you were suggesting, but thought that using the commit SHA made for a better looking API call than:

```
>install_gitlab("jimhester/covr@child_expressions")
Downloading GitLab repo jimhester/covr@child_expressions
from URL https://gitlab.com/api/v4/projects/jimhester%2Fcovr/repository/archive.tar.gz?sha=child_expressions
```

But that's not an important metric to maximise. I've reverted the commit and adopted your suggestions.
Thanks for your help.

jimhester:
Thanks for working on this!

@dpprdan Do the current changes look good to you?

@aornugent 

dpprdan:
@jimhester Looking good so far. 

However, the installation is not skipped, when `local_sha` is equal to `remote_sha`. The root of this is that `remote_package_name.gitlab_remote()` does not return a package name for non-public packages, because it does not use the Gitlab API (as was the case with `remote_download.gitlab_remote()`, originally). 

The consequence of this is that there is no `package_name` (also because `tryCatch()` does not fall back on `remote$repo`) -> there is no `local_sha` -> Installation cannot be skipped, because `local_sha` cannot be equal to `remote_sha`. 

See [this commit](https://github.com/dpprdan/remotes/commit/1481e26c8ba15eee6f8cd545e8a2f4632ec05a96) for the necessary changes. I can open a PR but you can also just copy it from there if you like, @aornugent. 

Other than that, everything looks fine to me, including installing from private repos with `@ref` or `/subdir`.

dpprdan:
@jimhester Two OT suggestions: 

1. Since the example is `install_gitlab("jimhester/covr")`, would you consider [mirroring](https://docs.gitlab.com/ce/workflow/repository_mirroring.html) the github repo to gitlab?
2. One thing I'd like to see on your Youtube series: How to debug a function that makes use of *apply/purrr::map*, e.g. via `debugonce()`. 😄

aornugent:
> See [this commit](https://github.com/dpprdan/remotes/commit/1481e26c8ba15eee6f8cd545e8a2f4632ec05a96) for the necessary changes. I can open a PR but you can also just copy it from there if you like, @aornugent.

Let's get it done in one.

jimhester:
Ok one last thing, can you run `make` in the package root so that the `install-github.R` script is updated? This should also fix the failing tests on the CI.

jimhester:
Thank you to @aornugent for the PR and for @dpprdan for improvements and reviewing the changes!

Thanks!! You are da bomb!

aornugent:
Oh no, we missed a bug here: https://github.com/r-lib/remotes/issues/355#issuecomment-490392168

This line

 https://github.com/r-lib/remotes/blob/92f30649ed90918ae66b074b270e2b2595493edb/R/install-gitlab.R#L122

doesn't need the access token at the end as it is added in the `download` call on line 125. I must've forgotten to test on a private repo.

What the best way to remedy this? A second PR or revert the merge?

aornugent:
Actually, I'm not sure that the double token is the cause of the bug, but I don't have a failing example to test on.

trafficonese:
In my case the double token is not the problem. I also get the "Not Found" error if I remove the second part. I dont know why, but our gitlab is not able to translate user/repo to project ID, with which it would work.

aornugent:
> I would fix it myself, but cannot get `make` to run successfully on Windows. 
I ended up installing WSL + Ubuntu + R to get it running. See: https://itsfoss.com/install-bash-on-windows/

dpprdan:
@aornugent Thanks! I do have `make` installed (part ot [git for windows](https://gitforwindows.org/)?), but I get an error while trying to run it.

<details>
<summary>Details</summary>

```bash

# First, what versions do I have?
$ make --version
GNU Make 4.2.1
Built for i686-pc-msys
Copyright (C) 1988-2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Daniel@Notebook-DP  ~/Documents/R/github/remotes (master)
$ R --version
R version 3.5.3 (2019-03-11) -- "Great Truth"
Copyright (C) 2019 The R Foundation for Statistical Computing
Platform: x86_64-w64-mingw32/x64 (64-bit)

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under the terms of the
GNU General Public License versions 2 or 3.
For more information about these matters see
http://www.gnu.org/licenses/.

# make cannot find the brew package
Daniel@Notebook-DP  ~/Documents/R/github/remotes (master)
$ make
Rscript -e 'brew::brew("inst/install-github.Rin", "inst/install-github.R")'
Error: object 'brew' not found

# But brew is installed?!
Daniel@Notebook-DP  ~/Documents/R/github/remotes (master)
$ Rscript -e 'library(brew);brew(system.file("example1.brew",package="brew"),envir=new.env())'
Title: brew test
current time: 2019-05-09 12:39:38
value of foo: bar
 i is 1
from example2.brew, foo is: bar1
 i is 2
from example2.brew, foo is: bar12
 i is 3
from example2.brew, foo is: bar123
 i is 4
from example2.brew, foo is: bar1234
 i is 5
from example2.brew, foo is: bar12345
 i is 6
from example2.brew, foo is: bar123456
 i is 7
from example2.brew, foo is: bar1234567
 i is 8
from example2.brew, foo is: bar12345678
 i is 9
from example2.brew, foo is: bar123456789
 i is 10
from example2.brew, foo is: bar12345678910

```

</details>

PR #363, opened 2019-05-08
Remove extra GitLab access token

aornugent:
Fixes small bug noted here: https://github.com/r-lib/remotes/issues/355#issuecomment-490392168

GitLab access token gets added to the API call in two places. This removes one.


jimhester:

Thanks!! You are da bomb!

dpprdan:
@jimhester 👍 
It might be worth mentioning #359 in NEWS as well - or instead of this, since this PR here fixes a bug in  #359 which was introduced after 2.0.4.

PR #367, opened 2019-05-15
Standalone JSON parser

gaborcsardi:
I haven't actually changed the code at all, it is just indented differently now.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,6 +26,7 @@ Imports:
 Suggests:
     brew,
     callr,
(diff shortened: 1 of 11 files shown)


gaborcsardi:
I'll some test that the file is indeed standalone.

gaborcsardi:
OK, I added a test via codetools, I checked that it works well, and does report globals, even if you reference to other functions in the package.

gaborcsardi:
Will merge once the CIs are clear.

PR #369, opened 2019-05-20
Forward type

krlmlr:
from `install_deps()` to `update()` .

Would it be better to set the corresponding option in the top-level function instead, to avoid passing `type` around?


jimhester:
Thanks!

PR #372, opened 2019-05-29
Add Bioconductor standalone file

gaborcsardi:
This is basically https://github.com/r-lib/pkgcache/pull/24, without any changes.

I will add some docs at the top of the file, otherwise it is ready, I think. 

remotes will contain the master copy of this file.

@jimhester @kevinushey Feedback welcome.


R/bioc-standalone.R
```diff
@@ -0,0 +1,217 @@
...
+    }
+    mirror <- Sys.getenv("R_BIOC_MIRROR", "https://bioconductor.org")
+    mirror <- getOption("BioC_mirror", mirror)
+    repos <- c(
+      BioCsoft      = "{mirror}/packages/{bv}/bioc",
+      BioCann       = "{mirror}/packages/{bv}/data/annotation",
+      BioCexp       = "{mirror}/packages/{bv}/data/experiment",
+      BioCworkflows =
+        if (bioc_version >= "3.7") "{mirror}/packages/{bv}/workflows",
+      BioCextra     =
```
kevinushey: Is there any way for us to detect if the set of BioConductor repositories changes in new releases? Or will we have to update this list as chose changes happen?

gaborcsardi: Good question! Unfortunately, AFAICT we can't easily automate this. The yaml file has the repos for the current devel version:
```yaml
## CHANGE THIS (i.e., uncomment) as various parts of the new devel version 
## become available. set to "[]" if none are available.
devel_repos:
- "bioc"
- "data/experiment"
- "data/annotation"
- "workflows"
```
but this is all, no repos for older versions, etc.

BiocManager has them hard-coded: https://github.com/Bioconductor/BiocManager/blob/6f8252609ddb9d2e02efbff96267ffb195ce1f32/R/repositories.R#L35-L36

There is also `tools:::.get_standard_repository_URLs()`, but that does not even have the correct repos now, and it is not updated often enough, anyway.
```r
❯ tools:::.get_standard_repository_URLs()
                                                   CRAN
                           "https://CRAN.R-project.org"
                                               BioCsoft
           "https://bioconductor.org/packages/3.9/bioc"
                                                BioCann
"https://bioconductor.org/packages/3.9/data/annotation"
                                                BioCexp
"https://bioconductor.org/packages/3.9/data/experiment"
```

Is there anything else I am missing?

What we could do is adding a test case that checks that (for R-devel/Bioc-devel?) we do have the correct set of repos, using the BiocManager package, which we would suggest. Then at least we will notice early if the repos change.

kevinushey:
Do we have any sense of whether the Bioconductor `config.yaml` is intended to be treated as a stable 'source of truth' for the state of Bioconductor repositories? Or is there some chance that it could be changed / replaced in the future?

In other words, I'm a bit worried that Bioconductor might decide that `BiocManager` is the only stable interface to Bioconductor package installation, and depending on some of these other external bits could burn us in the future.

gaborcsardi:
Another very valid concern.

As I understand, Bioc heavily relies on `config.yaml` as a source of truth, not just for Bioc versions, but also their builder infrastructure. BiocManager uses it as well. They advised us in favor of using this file, several times, in remotes and R-hub issues.

This PR actually depends on the YAML only mildly. In particular, it does not use it for R versions older than 3.6.x and it does not use it for 3.6.x either before October 2019. (Because it is unlikely that a new Bioc version comes out before that). Once the new Bioc version comes out, we update our pre-stored mappings. (And all copies of this file. :()

So if somebody uses an outdated version of remotes (renv, etc.), they could only get into trouble if they are using a newer version of R, for which the YAML will be consulted, _and_ if the Bioc team changes the YAML format or moves away from it.

I don't know any better source of truth, either. (But FIXME!)

I think we can add some test cases to alert ourselves:
- a test case that checks against CRAN BiocManager repos and our repos.
- a test case that simply fails if there is a Bioc release out, forcing us to update the package.

I should also add general test cases, anyway, to make sure that the YAML parsing still works against the current YAML file, etc. We are having some difficulties using covr for these standalone files, but we'll come up with a solution.

gaborcsardi:
@kevinushey @jimhester I added test cases.

jimhester:
@LiNk-NY I agree with Gabor and think the yaml file is the best place to dynamically query for Bioconductor repository versions, do you agree? (Because remotes by design has no dependencies we cannot simply rely on BiocManager directly to query this).

R/bioc-standalone.R
```diff
@@ -0,0 +1,217 @@
...
+      status <- rep("out-of-date", length(bioc))
+      release <- get_release_version()
+      devel <- get_devel_version()
+      status[bioc == release] <- "release"
+      status[bioc == devel] <- "devel"
+
+      # append final version for 'devel' R
+      bioc <- c(
+        bioc, max(bioc)
+        # package_version(paste(unlist(max(bioc)) + 0:1, collapse = "."))
```
jimhester: Should we remove this code now?

R/bioc-standalone.R
jimhester: I don't think there is anywhere else and in general the types of repos changes quite infrequently, so it shouldn't be too much work to keep track of manually.

install-github.R
```diff
@@ -3,124 +3,274 @@ function(...) {
...
+    version_map
   }
 
-  if (is.null(a)) {
-    p <- file.path(R.home("etc"), "repositories")
-    a <- ("tools" %:::% ".read_repositories")(p)
+  get_matching_bioc_version <- function(r_version = getRversion(),
+                                        forget = FALSE) {
+
+    minor <- get_minor_r_version(r_version)
```
jimhester: Any reason you don't use `r_version[, 1:2]` for this?

gaborcsardi:
The other issue with BiocManager is that it requires R 3.5, and we need to support older R versions.

gaborcsardi:
So I'll have to remove BiocManager from Suggests, actually, because of its R 3.5.0 version requirement. But will install it manually on Travis, and still have the test case there.

jimhester:
I think you can actually leave BiocManager in Suggests, but you then also need to set `_R_CHECK_FORCE_SUGGESTS_=false`.

gaborcsardi:
@jimhester I decided against that, to make the check clean on older R versions as well, even though on CRAN it would be OK. This BiocManager test case is really only for us, and we'll see the failures on Travis.

R/bioc-standalone.R
```diff
@@ -0,0 +1,247 @@
...
+  release_version <- NULL
+  version_map <- NULL
+  yaml_config <- NULL
+
+  # -------------------------------------------------------------------
+  # API
+
+  get_yaml_config <- function(forget = FALSE) {
+    if (forget || is.null(yaml_config)) {
+      new <- tryCatch(readLines(config_url), error = function(x) x)
```
kevinushey: How does `readLines()` handle URLs -- does it just invoke the moral equivalent of `utils::download.file()`? (just in case a user relies on e.g. `curl` + having things like `http_proxy` set)

kevinushey:
> They advised us in favor of using this file, several times, in remotes and R-hub issues.

That's good to know; thanks for confirming! This PR looks good to me then.

mtmorgan:
just skimming so might have missed things, but yes config.yaml is in many ways the underlying ground truth, for instance the info there is used to construct package landing pages where R isn't available.

The next version of BiocManager https://github.com/Bioconductor/BiocManager/blob/master/DESCRIPTION , which should be on CRAN soon (there is an issue with winbuilder that is a bit perplexing and requiring some careful investigation) does not Depend: on R >= 3.5, but does not do anything useful if used in an earlier version.

gaborcsardi:
@mtmorgan Thanks much for looking over this, and confirming!

R/bioc-standalone.R
gaborcsardi: Good point. It calls `file()` internally, which uses the `url.default` option. We'll have a "downloader" standalone file soon, and then this bioc file will use that. Would you prefer to use `download.file()` directly until then, instead of `file()`? I'll start writing the downloader code soon, buy it won't make it in the next remotes release.

kevinushey: I think `download.file()` would be a better choice for now.

gaborcsardi: OK, I'll change it to that.

gaborcsardi:
@kevinushey changed to `download.file()`. Added some docs as well. 

I am not yet sure what to do about the test coverage, we'll see in covr I guess.

@kevinushey @jimhester I'll wait for the CI to finish and then merge this, unless you want to change sg still.

gaborcsardi:
Thanks everyone for the comments!

PR #375, opened 2019-06-03
Fix is.na() warning in combine_deps for R 3.4 and below

### LINKED ISSUE #362: Warning in is.na(remote_deps$package): is.na() applied to non-(list or vector) of type 'NULL'
I'm getting the warning:
```
Warning in is.na(remote_deps$package): is.na() applied to non-(list or vector) of type 'NULL'
```
While installing packages with remotes 2.0.4 installed. This is caused by: https://github.com/r-lib/remotes/blob/91c64b3598fc8145bd20484193c68fa9c90bce04/R/deps.R#L147

And returning `NULL` here: https://github.com/r-lib/remotes/blob/91c64b3598fc8145bd20484193c68fa9c90bce04/R/deps.R#L506

Probably by this commit: https://github.com/r-lib/remotes/commit/8419037064bfcd3eccdca3e457d73b6a19357675

Maybe it just needs an `is.null(remote_deps)` check before...?

This is disruptive to us since we are running devtools with `options(warn = 2)` so that it correctly fails on actual errors that are reported as warnings. (Thank you, the R error handling mechanism... 😔)

cc @jimhester

segevfiner:
Not an R user. So please make sure this is correct. e.g. `remote_deps$package` can't be `NA` so we also need a `NA` check, etc.

Fixes #362

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -144,7 +144,7 @@ combine_deps <- function(cran_deps, remote_deps) {
   }
 
   # Only keep the remotes that are specified in the cran_deps or are NA
-  remote_deps <- remote_deps[is.na(remote_deps$package) | remote_deps$package %in% cran_deps$package, ]
+  remote_deps <- remote_deps[is.null(remote_deps$package) | remote_deps$package %in% cran_deps$package, ]
 
   # If there are remote deps remove the equivalent CRAN deps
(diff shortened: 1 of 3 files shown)


jimhester:
Thanks for the PR! I ended up fixing this in a different way, but it should no longer occur.

PR #391, opened 2019-06-24
Send auth information in the headers

### LINKED ISSUE #364: Send GH tokens in headers, instead of query parameters
With the user agent hack before 3.6.0.

To avoid exposing them when the URL is printed in error messages, etc.

gaborcsardi:
Instead of sending it in query parameters.
This way it is not printed in the logs on errors.

Also refactored the download code a bit. We still respect
the user's `download.file.method` option.

We have tests now for the various download methods.

Closes #364.


gaborcsardi:
@jimhester do not merge yet pls, tests on Windows are failing....

gaborcsardi:
OK, there is a slight problem. We cannot send headers with the wininet method on R versions older than R 3.6.0. Apparently the user agent trick does not work there.

gaborcsardi:
Since 3.2.x the Windows builds have libcurl, so this is not a major issue. I will add a warning if the user selected "wininet" on an older version of R _and_ headers are needed.

gaborcsardi:
Actually, we always require the curl package if R is older than 3.2.0, so this hardly ever comes up. Still, I'll add a warning for this rare case.

gaborcsardi:
OK, I think this is ready. The download code is pretty clean now.

jimhester:
Cool, looks good, thanks for working on this!

PR #397, opened 2019-07-09
Mention package name in RemoteType error

flying-sheep:

PR #406, opened 2019-07-24
package2remote: improve error msg

dpprdan:
This is a small fix which improves `package2remote`'s error message. In case it cannot convert the `RemoteType` it now also mentions which package is causing the error.


jimhester:
I just merged https://github.com/r-lib/remotes/pull/397 which does basically the same (although looking at this one that one forgot to update the standalone script)

Issue #396, opened 2019-07-08
`install_github` with non-default host broken: "attempt to set an attribute on NULL"


r-cheologist:
Consider the following:

    requireNamespace("remotes")
    remotes::install_github(
      repo = 'loosolab/MARMoSET', host = 'github.molgen.mpg.de/api/v3')

This fails with the error:

    Error: Failed to install 'MARMoSET' from GitHub:
      attempt to set an attribute on NULL

This happens both with the current cran version and after invoking `remotes::install_github("r-lib/remotes")`.

dpprdan:
The error message is caused by `ratelimit_reset <- .POSIXct(res_headers$`x-ratelimit-reset`, tz = "UTC")` in `github_error` since this server apparently does not return a `x-ratelimit-reset` in its header, so ``res_headers$`x-ratelimit-reset`` = NULL`. I gather that for this server the REST API rate limit is not set. In any case, this seems to be possible according to the docs 

> Note that the limits mentioned above are the default rate limits for a GitHub Enterprise Server instance. Contact your site administrator to confirm _if rate limits are enabled_ and how they are configured.

Emphasis mine. 

With my attempt to fix this applied (#413), we now get

``` r
remotes::install_github(
  repo = "loosolab/MARMoSET", host = "github.molgen.mpg.de/api/v3"
)
#> Error: Failed to install 'unknown package' from GitHub:
#>   HTTP error 401.
#>   Bad credentials
```

This is because `install_github()` uses the `auth_token = github_pat()` default.

Unfortunately `auth_token = NULL` does not help (yet).

``` r
remotes::install_github(
  repo = "loosolab/MARMoSET", host = "github.molgen.mpg.de/api/v3", 
  auth_token = NULL
)
#> Error: Failed to install 'unknown package' from GitHub:
#>   HTTP error 401.
#>   Bad credentials
```

[...]

PR #413, opened 2019-07-31
fix github_error() for when API does not return ratelimit

dpprdan:
A Github (Enterprise) instance does not always have rate-limiting activated and the corresponding API therefore does not return a `x-ratelimit-reset`, see #396.

This fixes `github_error()` for when this is the case.

Without this fix, the error message for the query in #396 is:

```
Error: Failed to install 'unknown package' from GitHub:
  attempt to set an attribute on NULL
```

Now it is:

```
Error: Failed to install 'unknown package' from GitHub:
  HTTP error 401.
  Bad credentials
```

(The package is unknown, because `github_error()` is triggered by `github_DESCRIPTION()`.)

I guess one could also add some `guidance` for 401 errors? Let me know if you want me too.

I also wondered whether the ratelimit information is indeed useful for non-429 HTTP errors? What do I do with that information, when it isn't causing the problem? If that info can be omitted, that could make the code cleaner, namely these three or four cases:

1. 404 error
1. possibly 401 error
2. 429 error (+ ratelimit info)
3. all other errors.

Please let me know what you think.


jimhester:
How about

```r
res_headers$`x-ratelimit-reset` %||% NA
```

This way you avoid needing any additional conditionals `format(.POSIXct(NA), usetz = TRUE)` is `NA`, so it should be fine.

Seeing the rate limit information is useful even if it is not a 429 error, it lets me know if someone is using authenticated requests or not, (the limit is only 60 for unauthenticated requests).

dpprdan:
`%||%` is better indeed. 

I still left the last conditional in there, because IMHO the rate limit information in 

```r
Error: Failed to install 'MARMoSET' from GitHub:
  HTTP error 401.
  Bad credentials

  Rate limit remaining: NA/NA
  Rate limit reset at: NA
```

is potentially misleading from a _user_ perspective (i.e. suggesting that something is wrong with the rate limit while the problem is the credentials). So right now the two rate limit lines won't be displayed (see my original post). But please let me know, if you want me to change that.

jimhester:
Cool, looks much cleaner now, thanks! 
Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`, then I can merge this PR.

dpprdan:
👍 done!

jimhester:
Thanks!

Issue #361, opened 2019-05-06
install_bioc() not install dependecies from same Bioconductor version


dongzhuoer:
Current Bioconductor release is 3.9, on which latest [beachmat](http://bioconductor.org/packages/release/bioc/html/beachmat.html) is 2.0.0.

But when I `remotes::install_bioc('release/BiocSingular')`, it use [1.4.0](http://bioconductor.org/packages/3.8/bioc/html/beachmat.html), which belongs to Bioconductor 3.8. 

It's the same for other packages, like BiocGenerics. I guess all the dependencies are installed from Bioconductor 3.8.

```
R -e "remotes::install_bioc('release/BiocSingular')"

R version 3.6.0 (2019-04-26) -- "Planting of a Tree"
Copyright (C) 2019 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu (64-bit)

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

> remotes::install_bioc('release/BiocSingular')
Downloading Bioconductor repo https://git.bioconductor.org/packages/BiocSingular
'/usr/bin/git' clone --depth 1 --no-hardlinks --branch RELEASE_3_9 https://git.bioconductor.org/packages/BiocSingular /tmp/RtmpSU2ryC/file1577c9f63
Cloning into '/tmp/RtmpSU2ryC/file1577c9f63'...
beachmat     (NA -> 1.4.0   ) [CRAN]
BH           (NA -> 1.69.0-1) [CRAN]
BiocGenerics (NA -> 0.28.0  ) [CRAN]
BiocParallel (NA -> 1.16.6  ) [CRAN]
DelayedArray (NA -> 0.8.0   ) [CRAN]
formatR      (NA -> 1.6     ) [CRAN]
futile.lo... (NA -> 1.4.3   ) [CRAN]
futile.op... (NA -> 1.0.1   ) [CRAN]
HDF5Array    (NA -> 1.10.1  ) [CRAN]
IRanges      (NA -> 2.16.0  ) [CRAN]
[...]

PR #415, opened 2019-08-02
Handle Bioconductor-installed packages

flying-sheep:
This uses the DESCRIPTION fields set by `BiocManager::install` to create `bioc_xgit` remotes instead of misidentifying those as CRAN-installed packages.

It would maybe be good to reintroduce the `bioc` remote type for packages installed by `BiocManager::install`.

### DIFF
--- a/R/install-remote.R
+++ b/R/install-remote.R
@@ -191,8 +191,22 @@ package2remote <- function(name, lib = .libPaths(), repos = getOption("repos"), 
         sha = NA_character_))
   }
 
-  if (is.null(x$RemoteType) || x$RemoteType == "cran") {
+  # BiocManager sets a few fields we can use. Does not work for AnnotationDBI packages like org.Hs.eg.db.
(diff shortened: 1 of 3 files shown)


R/install-remote.R
```diff
@@ -191,8 +191,22 @@ package2remote <- function(name, lib = .libPaths(), repos = getOption("repos"),
         sha = NA_character_))
   }
 
-  if (is.null(x$RemoteType) || x$RemoteType == "cran") {
+  # BiocManager sets a few fields we can use. Does not work for AnnotationDBI packages like org.Hs.eg.db.
+  if (is.null(x$RemoteType) && !is.null(x$git_url) && requireNamespace("BiocManager", quietly = TRUE)) {
+    bioc_pkgs <- rownames(available.packages(repos = BiocManager::repositories()))
```
flying-sheep: This is a cached operation, so it won’t be slow.

jimhester:
Thank you for opening the issue and for working on the pull request. However as I mentioned in #361 I believe this is working as intended already.

Even though we're closing this pull request, we really appreciate the feedback, and hope you'll continue to contribute in the future :smile:

Issue #411, opened 2019-07-29
Dependencies install to the wrong place


kenahoo:
When doing `remotes::install_version(pkg, ver, lib=location)`, the package `pkg` will get installed to the correct location `location`, but any dependencies will not.

This is because `remotes::install` doesn't pass along `...` to the `install_deps()` call:

```
> packageVersion('remotes')
[1] '2.1.0'
> remotes:::install
function (pkgdir, dependencies, quiet, build, build_opts, build_manual, 
    build_vignettes, upgrade, repos, type, ...) 
{
    warn_for_potential_errors()
    if (file.exists(file.path(pkgdir, "src"))) {
        if (has_package("pkgbuild")) {
            pkgbuild::local_build_tools(required = TRUE)
        }
        else if (!has_devel()) {
            missing_devel_warning(pkgdir)
        }
    }
    pkg_name <- load_pkg_description(pkgdir)$package
    if (is_root_install()) 
        on.exit(exit_from_root_install(), add = TRUE)
    if (check_for_circular_dependencies(pkgdir, quiet)) {
        return(invisible(pkg_name))
    }
    install_deps(pkgdir, dependencies = dependencies, quiet = quiet, 
        build = build, build_opts = build_opts, build_manual = build_manual, 
        build_vignettes = build_vignettes, upgrade = upgrade, 
        repos = repos, type = type)
    if (isTRUE(build)) {
        dir <- tempfile()
        dir.create(dir)
        on.exit(unlink(dir), add = TRUE)
        pkgdir <- safe_build_package(pkgdir, build_opts, build_manual, 
            build_vignettes, dir, quiet)
    }
    safe_install_packages(pkgdir, repos = NULL, quiet = quiet, 
        type = "source", ...)
    invisible(pkg_name)
}
<bytecode: 0x1fd4020>
<environment: namespace:remotes>
```

kenahoo:
Also - I assume this affects more functions than just `install_version`, since it looks like a problem with the `install()` workhorse function, right?  Would affect all the `install_*` functions, basically if I'm not mistaken.

tannerntannern:
I'm also experiencing this issue.  It's quite a problem because I need to install some packages to an external directory to zip them up and use them on another platform, but the dependencies all go to the default library directory, which means I have to manually reinstall those dependencies in the external folder, and the dependencies of those dependencies, etc...

jimhester:
This will be fixed once https://github.com/r-lib/remotes/pull/402 is merged.

[...]

PR #444, opened 2019-10-04
Pass ellipsis parameters when install dependencies

Neil-Schneider:
This PR is related to #411. I ran across this when trying to install packages in an instance without docs, so the INSTALL_opts were not propagating. I fixed this before I realized that #402 exists but that PR seems to have stalled.


jimhester:

Thanks!! You are da bomb!

antoine-sachet:
Something more complex may be needed. Passing all the `...` to dependencies has unexpected consequences (remember #345 which I "fixed" by removing the `...` from `install_deps`).

PR #450, opened 2019-10-22
be resilient to missing repos / pkg_type

kevinushey:
While this bug was originally surfaced by a bad behavior in `renv`, I think it doesn't hurt to be resilient in case these cases are encountered.


jimhester:
Could we instead put the fallbacks in https://github.com/r-lib/remotes/blob/5b3da5f852772159cc2a580888e10536a43e9199/R/install-remote.R#L207-L208? I think that would also solve the issue

kevinushey:
Good idea! I've made that change.

jimhester:
Awesome, great work!

Thanks a bunch!

kevinushey:
😄 Thank you!

Issue #459, opened 2019-12-03
Support declaring extra dependencies


jeroen:
As discussed on slack, we want to formalize declaring several types of development dependencies in the description file. These are packages that are _not_ needed to pass check, but used  for development. E.g:

 - pkgdown: packages used by pkgdown articles
 - document: packages such as roxygen2 needed to generate manuals

By declaring these, we can automate rendering docs without having to do custom stuff in the travis file, or making these packages a requirement for running CMD check (as Suggests does).

The proposal is to prefix the fields with `Config/Depends/` such that they do not trigger a NOTE.

```txt
Imports: foo, bar
Suggest: baz
Config/Depends/pkgdown: tidyverse, bench
Config/Depends/coverage: covr
Config/Depends/document: roxygen2
```

Fields may contain remotes syntax like `cran::dplyr@0.8.2, r-lib/roxygen@branch`.

We generalize the `dependencies` parameter from the `remotes::install_...()` functions to support these custom dependencies such as `"pkgdown"`.

One open question is the dependency relationship between the dependency types :) For example:

```r
remotes::install_github("dplyr", dependencies = "pkgdown")
```

This should probably dependencies in `Config/Depends/pkgdown` plus hard dependencies (Depends, Imports, LinkingTo), but not maybe Suggests (which are mostly required to run tests)?

gaborcsardi:
I think `"pkgdown"` should only install `Config/Depends/pkgdown`, to be able to express everything. In general, if it is a character vector, then it should only install the specified stuff.

So for your site you would typically do `c("hard", "soft", "pkgdown")`.

We can also have `"all"`, to install absolutely everything (non-recursively of course), and `TRUE` would mean `c("hard", "soft")`, like it already does.

maelle:
I was wondering why `usethis::use_coverage()` added covr to Suggests (even if covr is only needed on Travis), and I suppose the proposal by Jeroen would mean that could be avoided?

hadley:
I wonder if `Config/Needs/...` would be slightly better than `Config/Depends/...` since it avoids the existing meaning of the "Depends" field.

jimhester:
Closed by https://github.com/r-lib/remotes/pull/483

PR #483, opened 2020-03-17
Support declaring extra dependencies

jimhester:
This allows remotes to install extra dependencies if requested, which is
useful for things like declaring pkgdown and coverage dependencies,
among others.


jimhester:
This seems to be sufficient changes for remotes to work with arbitrary fields. Rather than always prefixing with `Config/` I just have it pass the full field name to `dependencies`, though in practice I think we will generally use `Config/pkgdown` or similar fields.

jeroen:
Testing this, I'm getting:

```r
> remotes::install_github('ropensci/av', dependencies = 'Config/pkgdown')
Downloading GitHub repo ropensci/av@master
These packages have more recent versions available.
It is recommended to update all of them.
Which would you like to update?

1: All                            
2: CRAN packages only             
3: None                           
4: ggplot2 (3.2.1 -> 3.3.0) [CRAN]

Enter one or more numbers, or an empty line to skip updates:
2
phonTools (NA -> 0.2-2.1) [CRAN]
tuneR     (NA -> 1.3.3  ) [CRAN]
Installing 1 packages: phonTools
Error: Failed to install 'av' from GitHub:
  Failed to install 'phonTools' from CRAN:
  subscript out of bounds
```

jimhester:
Ok I see the same thing, it should be fixed after 2037314. Then running the following seems to work for 'ropensci/av'

```r
remotes::install_github("ropensci/av", dep = c("hard", "Config/pkgdown"))
```

jeroen:
Thanks. It works now, though it seems it is trying to install evertyhing twice?

```
> remotes::install_github("ropensci/av", dep = c("hard", "Config/pkgdown"))
Downloading GitHub repo ropensci/av@master
Installing 1 packages: phonTools
trying URL 'https://cran.rstudio.com/bin/macosx/el-capitan/contrib/3.6/phonTools_0.2-2.1.tgz'
Content type 'application/x-gzip' length 461275 bytes (450 KB)
==================================================
downloaded 450 KB


The downloaded binary packages are in
	/var/folders/l5/gk7g18z52876d4lbk9tjhr0c0000gn/T//Rtmp61Nylb/downloaded_packages
Installing 1 packages: signal
trying URL 'https://cran.rstudio.com/bin/macosx/el-capitan/contrib/3.6/signal_0.7-6.tgz'
Content type 'application/x-gzip' length 345167 bytes (337 KB)
==================================================
downloaded 337 KB


The downloaded binary packages are in
	/var/folders/l5/gk7g18z52876d4lbk9tjhr0c0000gn/T//Rtmp61Nylb/downloaded_packages
Installing 1 packages: tuneR
trying URL 'https://cran.rstudio.com/bin/macosx/el-capitan/contrib/3.6/tuneR_1.3.3.tgz'
Content type 'application/x-gzip' length 560777 bytes (547 KB)
==================================================
downloaded 547 KB


The downloaded binary packages are in
	/var/folders/l5/gk7g18z52876d4lbk9tjhr0c0000gn/T//Rtmp61Nylb/downloaded_packages
Skipping install of 'phonTools' from a cran remote, the SHA1 (0.2-2.1) has not changed since last install.
  Use `force = TRUE` to force installation
Skipping install of 'signal' from a cran remote, the SHA1 (0.7-6) has not changed since last install.
  Use `force = TRUE` to force installation
Skipping install of 'tuneR' from a cran remote, the SHA1 (1.3.3) has not changed since last install.
  Use `force = TRUE` to force installation
✓  checking for file ‘/private/var/folders/l5/gk7g18z52876d4lbk9tjhr0c0000gn/T/Rtmp61Nylb/remotese7491058ee46/ropensci-av-0d88785/DESCRIPTION’ ...
─  preparing ‘av’:
✓  checking DESCRIPTION meta-information ...
─  cleaning src
─  running ‘cleanup’
─  checking for LF line-endings in source and make files and shell scripts
─  checking for empty or unneeded directories
   Removed empty directory ‘av/vignettes’
─  building ‘av_0.5.0.tar.gz’
   
* installing *source* package ‘av’ ...
** using staged installation
Found pkg-config cflags and libs!
Using PKG_CFLAGS=-I/usr/local/Cellar/ffmpeg/4.2.2_2/include
Using PKG_LIBS=-L/usr/local/Cellar/ffmpeg/4.2.2_2/lib -lavfilter
** libs
rm -f av.so fft.o formats.o info.o init.o video.o winfunc.o
clang -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -I/usr/local/Cellar/ffmpeg/4.2.2_2/include  -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -I/usr/local/include  -fPIC  -Wall -g -O2  -c fft.c -o fft.o
clang -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -I/usr/local/Cellar/ffmpeg/4.2.2_2/include  -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -I/usr/local/include  -fPIC  -Wall -g -O2  -c formats.c -o formats.o
clang -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -I/usr/local/Cellar/ffmpeg/4.2.2_2/include  -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -I/usr/local/include  -fPIC  -Wall -g -O2  -c info.c -o info.o
clang -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -I/usr/local/Cellar/ffmpeg/4.2.2_2/include  -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -I/usr/local/include  -fPIC  -Wall -g -O2  -c init.c -o init.o
clang -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -I/usr/local/Cellar/ffmpeg/4.2.2_2/include  -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -I/usr/local/include  -fPIC  -Wall -g -O2  -c video.c -o video.o
clang -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -I/usr/local/Cellar/ffmpeg/4.2.2_2/include  -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -I/usr/local/include  -fPIC  -Wall -g -O2  -c winfunc.c -o winfunc.o
clang++ -std=gnu++11 -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress -L/Library/Frameworks/R.framework/Resources/lib -L/usr/local/lib -o av.so fft.o formats.o info.o init.o video.o winfunc.o -L/usr/local/Cellar/ffmpeg/4.2.2_2/lib -lavfilter -F/Library/Frameworks/R.framework/.. -framework R -Wl,-framework -Wl,CoreFoundation
installing to /Library/Frameworks/R.framework/Versions/3.6/Resources/library/00LOCK-av/00new/av/libs
** R
** inst
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
** building package indices
** testing if installed package can be loaded from temporary location
** checking absolute paths in shared objects and dynamic libraries
** testing if installed package can be loaded from final location
** testing if installed package keeps a record of temporary installation path
* DONE (av)
```

NEWS.md
```diff
@@ -1,6 +1,11 @@
 # remotes (development version)
 
 * Another fix for the mixed binary and source dependency issue, it should hopefully be fully squashed now (#296)
+* Remotes functions can now install dependencies from additional DESCRIPTION
+  fields, e.g. passing `dependencies = "Config/pkgdown"` will install the
```
hadley: Should this be `Config/remotes/...` to allow for use of `Config/` by other packages?

R/deps.R
```diff
@@ -16,6 +16,14 @@
 #'   "Suggests". `NA` is shorthand for "Depends", "Imports" and "LinkingTo"
 #'   and is the default. `FALSE` is shorthand for no dependencies (i.e.
 #'   just check this package, not its dependencies).
+#'
+#'   The values "soft" or "dev" mean the same as `TRUE`, "hard" or "user" mean
+#'   the same as `NA`.
+#'
+#'   You can also specify dependencies from one or more additional fields,
+#'   common ones are "Config/pkgdown" for dependencies used in building the
```
hadley: I wonder if we should use a more generic name, like "website"`? I'd also make this a bulleted list now, since we're likely to slowly add other fields.

hadley: Do we want to add `Config/Remote/development` now? (i.e. you'd put roxygen2 in this field)

jimhester: Was thinking of starting slow, with just things we need on CI systems and go from there...

gaborcsardi:
This closes #459.

R/deps.R
```diff
@@ -407,7 +429,13 @@ standardise_dep <- function(x) {
   } else if (identical(x, FALSE)) {
     character(0)
   } else if (is.character(x)) {
-    x
+    if (any(x %in% c("hard", "user"))) {
+      c("Depends", "Imports", "LinkingTo")
+    } else if (any(x %in% c("soft", "dev"))) {
+      c("Depends", "Imports", "LinkingTo", "Suggests")
```
gaborcsardi: `Enhances`?

R/deps.R
```diff
@@ -407,7 +429,13 @@ standardise_dep <- function(x) {
...
+    } else {
+      intersect(x, c("Depends", "Imports", "LinkingTo", "Suggests"))
```
gaborcsardi: `Enhances`?

NEWS.md
gaborcsardi: Or, actually `Config/Needs/...` seems even better, as per https://github.com/r-lib/remotes/issues/459#issuecomment-590498663

R/deps.R
jimhester: I was treating this like `dependencies = TRUE`, which does _not_ seem to include `Enhances`

NEWS.md
jimhester: The extra namespacing seems unneeded until there is a clash, and if there ever is a clash then we can add just start using it, the code would stay the same.

R/deps.R
```diff
@@ -407,7 +429,13 @@ standardise_dep <- function(x) {
   } else if (identical(x, FALSE)) {
     character(0)
   } else if (is.character(x)) {
-    x
+    if (any(x %in% c("hard", "user"))) {
+      c("Depends", "Imports", "LinkingTo")
+    } else if (any(x %in% c("soft", "dev"))) {
```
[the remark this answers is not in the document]
jimhester: Well then you would just pass `Config/Depends/devel` in that case...

NEWS.md
gaborcsardi: If we don't have an extra namespace, that means that we are claiming the whole `Config/` namespace for dependency purposes. E.g. if we install packages in `Config/pkgdown` when building the pkgdown site, that rules out for example pkgdown storing its own config in `Config/pkgdown`.

R/deps.R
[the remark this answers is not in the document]
gaborcsardi: But it would make sense to include `Config/Depends/devel` in `dev` by default. If you don't include it now, then you'll break the API when you add it later.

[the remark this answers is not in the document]
jimhester: Ok, I got rid of the user / dev options

NEWS.md
jimhester: Ok, I added a 'Config/needs/website' example in the examples

jimhester:
So should we use `Config/Needs/xyz` or `Config/needs/xyz`? We don't need the capital after `Config` to pass R CMD check without a note, and I am not sure which one users would expect...

gaborcsardi:
I am not sure, either. FWIW, similar fields are uppercase, e.g. `Date/Publication`: https://github.com/wch/r-source/blob/30b99ab81158175f953a9d22772d5968e34ff600/src/library/tools/R/utils.R#L1330

NEWS.md
gaborcsardi: Thanks!

jimhester:
I guess we can use `Config/Needs` then

NEWS.md
hadley: Given that we now know about it, I suspect the chances of using `Config` for other packages is high, and I think the most natural way to divide up the `Config` namespace is via package name (since that already provides a system for "claiming" a name)

jimhester: Oh, so you are saying `Config/remotes/website` is what this should be called.

My main hesitation is that I would guess in the future pak and renv may also use this field, but maybe that is ok.

jimhester: Also calling it `Config/remotes` may confuse people with the existing `Remotes:` field.

gaborcsardi: Yeah, I think `Config/Needs` is more readable, and renv and pak will also use it. I don't think `Config/remotes` is bad, though, I think that's also reasonable.

PR #486, opened 2020-03-22
Fix an issue in the R -> Bioc version mapping

gaborcsardi:
If the current R version is not in the builtin table, then the
map might contain an R version twice, and we need a more
sophisticated way to choose the right Bioc version.

### DIFF
--- a/R/bioc-standalone.R
+++ b/R/bioc-standalone.R
@@ -64,7 +64,8 @@
 #'
 #' \section{NEWS:}
 #' * 2019-05-30 First version in remotes.
-#'
+#' * 2020-03-22 get_matching_bioc_version() is now correct if the current
+#'              R version is not in the builtin mapping.
 #'
 #' @name bioconductor
 #' @keywords internal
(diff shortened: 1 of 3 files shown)


R/bioc-standalone.R
```diff
@@ -200,10 +201,24 @@ bioconductor <- local({
...
+    mine <- which(package_version(minor) == map$r_version)
+    if (length(mine) == 0) {
+      mine <- NA
+    } else if (length(mine) > 1) {
+      if ("release" %in% map$bioc_status[mine]) {
+        mine <- mine["release" == map$bioc_status[mine]]
+      } else if ("devel" %in% map$bioc_status[mine]) {
+        mine <- mine["devel" == map$bioc_status[mine]]
+      } else {
+        mine <- rev(mine)[1]
```
jimhester: Should this be `[[1]]`?

jimhester: Also maybe using `tail(mine, n = 1)` is possibly more clear?

gaborcsardi: No, it has to be `[1]` because we need an `NA` if there is no match, instead of an error.

We cannot use `tail()` because this is a standalone file, it should not import the utils package, ideally. Although we already load it for `download.file`, but still in general we should avoid non-base functions.

jimhester: Ok, but aren't we handling the no match case with `length(mine) == 0` above?

PR #489, opened 2020-03-30
Make install_*() abortable

krlmlr:
because it's notoriously difficult to abort otherwise, at least in RStudio.

I'm using devmode, CRAN packages go to the "stable" library, GitHub packages to the "dev" one. If an update is required, I never want to upgrade packages from CRAN (because they would land in "dev"), instead I want to upgrade "stable" manually before proceeding.

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -593,7 +593,13 @@ upgradable_packages <- function(x, upgrade, quiet, is_interactive = interactive(
         choices <- c("All", "CRAN packages only", "None", choices)
       }
 
+      choices <- c(choices, "Abort")
+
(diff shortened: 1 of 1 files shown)


jimhester:
This seems like a [bug in RStudio](https://github.com/rstudio/rstudio/issues/2184), on the terminal you can send an interrupt to abort it, but RStudio does not seem to support this when the R process is waiting for input.

krlmlr:
If the mountain won't come to the prophet...

jimhester:
Turns out RStudio supports interrupting using `readline()`, but not `readLines()`, so we can just switch to the former to get the same interrupt behavior in RStudio as on the console.

PR #493, opened 2020-04-19
vignette: update sample Bitbucket path

infotroph:
...because `lsr` has moved.

(I realize these are just for illustration, but having the examples all resolve to real URLs definitely helps me understand the syntax)


jimhester:
Thanks!

Issue #495, opened 2020-04-26
download_version doesn't look up duplicated package


kos59125:
Local CRAN repositories may have duplicated packages (same name with different versions). For example, if a repository owner uploads package PKG v1.0 and PKG 2.0 to their CRAN repository, PACKAGE in contriburl is like:

```
Package: PKG
Version: 1.0
Depends: anotherpkg
MD5sum: hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh
NeedsCompilation: no

Package: PKG
Version: 2.0
Depends: anotherpkg
MD5sum: hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh
NeedsCompilation: no
```

This is not abnormal as `tools::write_PACKAGES(latestOnly = FALSE)` can generate so.

The current version of `download_version_url` uses `available.packages` with default filters which means `c("R_version", "OS_type", "subarch", "duplicates")`. `"duplicates"` means it returns the latest version if duplicated packages are found (see `help(available.packages)`).

There are two problems. The first problem is, if users call `download_version("PKG", "1.0", "https://my_local_repo")` with default filters, it raises an error (if there's no Archive directory like official CRAN) because the latest version of PKG is not 1.0 but 2.0. The second problem is, if users set the `available_packages_filters` option like `c("R_version", "OS_type", "subarch")`, the user cannot download the latest version with `download_version("PKG", "2.0", "https://my_local_repo")`.

My suggestion to resolve the problems is that `download_version_url` should consider the `available_packages_filters` option and remove `"duplicates"` filter. If users do not set the `available_packages_filters` option, the function use `c("R_version", "OS_type", "subarch")`. If users do set the `available_packages_filters` option, the function uses `setdiff(user_filters, "duplicates")`.

jimhester:
Could you see if https://github.com/r-lib/remotes/pull/305 fixes your issue, I think it may.

kos59125:
It partially works: `install_version("PKG", "1.0", repos = ...)` installs v1.0 and `install_version("PKG", "2.0", repos = ...)` installs v2.0, but `install_version("PKG", repos = ...)` (version is NULL) doesn't install the latest version.

kos59125:
I thought `version = NULL` will download the latest version (as documented in the current version of `remotes::download_version`), but I found the HEAD version doesn't specify the behavior of `version = NULL`.

PR #496, opened 2020-04-28
Handle available packages filters

kos59125:
The PR fixes the problem described in #495.


kos59125:
The test results say that tests I added succeed on Linux, but failed on Windows. I don't know the fundamental reason, but it seems that Windows doesn't work against local CRAN URL with scheme `file:///`. I added the local repo tests because I don't know such public repositories with duplicated packages as I described in #495.

I would like to skip tests which I added on Windows if the package owner allows. Otherwise, any ideas to fix the tests?

Issue #508, opened 2020-06-13
Add support for alternative default branches in GitHub


ha0ye:
`remotes::install_github()` assumes that the default branch is named "master", which is not always the case, e.g.
``` r
remotes::install_github("elizagrames/litsearchr")
#> Using github PAT from envvar GITHUB_PAT
#> Downloading GitHub repo elizagrames/litsearchr@master
#> Error in utils::download.file(url, path, method = method, quiet = quiet,  : 
#>   cannot open URL 'https://api.github.com/repos/elizagrames/litsearchr/tarball/master'
```

<sup>Created on 2020-06-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

The GitHub API notes a `default_branch` field that could provide this information, e.g.:
``` r
remotes:::github_GET("repos/elizagrames/litsearchr")$default_branch
#> [1] "main"
```

<sup>Created on 2020-06-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

So, I think this can be a drop-in for the default value for the `ref` argument in `install_github`. Some additional re-factoring might be needed if we wanted to bundle the GitHub API calls.

Note, to my knowledge, this issue was first reported by @elizagrames [on twitter](https://twitter.com/ElizaGrames/status/1271870736496541696?s=20), and we chatted about a potential fix before writing up this issue. We're happy to work on a PR to this, and look into propagating the solution to other `install_*` functions, if so desired.

jimhester:
I think this would be a non-starter unless you can implement it only using the existing number of GitHub API calls, we already run into API limiting problems with the current number of calls, so we can't really do an additional call to lookup the default branch.

If you can work out a way to do this reusing the same number of API calls we could probably review a pull request for it.

gaborcsardi:
Might be possible with the graphql api.

MyKo101:
I don't know how far reaching this is inside the inner workings of `remotes` (i.e. in other functions), but I can't see this being hard to implement. Forgive me if it is more complicated (particularly with sources other than github), but to me it seems straightforward.

Running `install_github()` with `ref=""` finds the default branch automatically regardless of it's name.

`remotes:::remote_download.github_remote()` builds a url for the desired tarball directory and then downloads it. A references can be tacked on at the end of the url, but if you're downloading the default branch, you don't need it.

For example:
[...]

PR #510, opened 2020-06-18
Set default branch from "master" to "HEAD"

MyKo101:
This PR replaces `"master"` within reference to repo branches with `"HEAD"`, which grabs the latest commit from the `default` branch. This allows for repos that do not use the `"master"` convention to be targeted for download without the need to specify a reference (e.g. `"main"`).

This will not effect repos that already use `"master"` as their default branch, as it will still be pulled. The only situation negatively effected by this PR is when a user is trying to target the `"master"` branch (and is thus using the previous default, `ref="master"`) and it is not the default, which will be a rare edgecase.

It is also suggested that the `remotes` repository changes it's default branch to `"main"`. Currently codecov requires a named branch and is incompatible with `"HEAD"`, and so the badge (as it stands in this PR) will not display until this is done.

This resolves Issue #508


README.md
```diff
@@ -7,11 +7,11 @@
 
 <!-- badges: start -->
 [![Project Status: Active - The project has reached a stable, usable state and is being actively developed.](http://www.repostatus.org/badges/latest/active.svg)](http://www.repostatus.org/#active)
-[![Linux Build Status](https://travis-ci.org/r-lib/remotes.svg?branch=master)](https://travis-ci.org/r-lib/remotes)
+[![Linux Build Status](https://travis-ci.org/r-lib/remotes.svg)](https://travis-ci.org/r-lib/remotes)
 [![Windows Build status](https://ci.appveyor.com/api/projects/status/github/r-lib/remotes?svg=true)](https://ci.appveyor.com/project/gaborcsardi/remotes)
 [![](https://www.r-pkg.org/badges/version/remotes)](https://www.r-pkg.org/pkg/remotes)
 [![CRAN RStudio mirror downloads](https://cranlogs.r-pkg.org/badges/remotes)](https://www.r-pkg.org/pkg/remotes)
-[![Coverage Status](https://img.shields.io/codecov/c/github/r-lib/remotes/master.svg)](https://codecov.io/github/r-lib/remotes?branch=master)
```
jimhester, line 14: You need to keep `?branch=xyz` here and for the travis case, otherwise the badge would show the status of the last running branch instead of the default one.

jimhester:
While I agree that a change from 'master' as the default branch of remotes would be preferable, I think we should wait to change the default branches on this and other repos until GitHub or git officially changes to a different default branch name. Right now the only 'official' word on this from GitHub is a twitter reply (https://twitter.com/natfriedman/status/1271253144442253312), and this change will likely break a number of workflows, so it needs to be done carefully.

For now could you update this PR to keep the default branch for remotes as 'master', and add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

MyKo101:
I understand the apprehension, as discussed on the Issue, some people's workflows will be largely affected. However, I believe it useful to have this ready to go and at some point this change will be needed.

I've added to the NEWS.md file as requested. This is actually my first PR (other than in my own repos), so any feedback on my style, etc... would be greatly appreciated.

NEWS.md
```diff
@@ -1,7 +1,10 @@
 # remotes (development version)
 
-* Another fix for the mixed binary and source dependency issue, it should hopefully be fully squashed now (#296)
-* The upgrade menu is now interruptible in RStudio (#489)
+* `install_*()` family of functions defaults to the `default` branch,
```
gaborcsardi, line 3:
```suggestion
* `install_*()` family of functions defaults to the default branch, 
```

R/install-github.R
```diff
@@ -9,7 +9,7 @@
 #'   (see below); if both is specified, the values in `repo` take
 #'   precedence.
 #' @param ref Desired git reference. Could be a commit, tag, or branch
-#'   name, or a call to [github_pull()]. Defaults to `"master"`.
+#'   name, or a call to [github_pull()]. Defaults to `"HEAD"`.
```
gaborcsardi, line 12:
```suggestion
#'   name, or a call to [github_pull()]. Defaults to `"HEAD"`, which means the default branch on GitHub.
```

R/submodule.R
```diff
@@ -79,7 +79,8 @@ update_submodule <- function(url, path, branch, quiet) {
   if (length(branch) > 0 && !is.na(branch)) {
     args <- c(args, "--branch", branch)
   }
-  args <- c(args, url, path)
+  #args <- c(args,paste0("'",url,"'"),paste0("'", path,"'"))
```
gaborcsardi, line 82: Can you pls remove this?

R/submodule.R
```diff
@@ -79,7 +79,8 @@ update_submodule <- function(url, path, branch, quiet) {
...
+  args <- c(args,url,path)
```
gaborcsardi, line 83: Can you pls put spaces after the commas? Thanks.

README.md
```diff
@@ -52,7 +52,7 @@ install.packages("remotes")
 Note that most of the examples here use GitHub. See below for other
 supported repository types.
 
-To install the latest version of a package in the `master` branch from
+To install the latest version of a package in the `default` branch from
```
gaborcsardi, line 55:
```suggestion
To install the latest version of a package in the default branch from
```

tests/testthat/test-json.R
```diff
@@ -81,7 +81,7 @@ test_that("JSON parser, real examples", {
       "email": "h.wickham@gmail.com",
       "date": "2015-03-30T13:55:18Z"
     },
-    "message": "Merge pull request #22 from paulstaab/master\\n\\nImprove error message for assertions of length 0",
+    "message": "Merge pull request #22 from paulstaab/HEAD\\n\\nImprove error message for assertions of length 0",
```
gaborcsardi, line 84: You cannot change this I believe, or the tests will fail.

tests/testthat/test-json.R
```diff
@@ -102,7 +102,7 @@ test_that("JSON parser, real examples", {
         name = "Hadley Wickham",
         email = "h.wickham@gmail.com",
         date = "2015-03-30T13:55:18Z"),
-      message = "Merge pull request #22 from paulstaab/master\\n\\nImprove error message for assertions of length 0",
+      message = "Merge pull request #22 from paulstaab/HEAD\\n\\nImprove error message for assertions of length 0",
```
gaborcsardi, line 105: You cannot change this one, either.

tests/testthat/test-script.R
```diff
@@ -61,3 +61,4 @@ test_that("install-github.R script does not load any package", {
 
   expect_equal(pkgs[[1]], pkgs[[2]])
 })
+
```
gaborcsardi, line 64: Can you pls remove the whitespace change?

tests/testthat/test-submodule.R
```diff
@@ -93,43 +93,47 @@ test_that("Can install a repo with a submodule", {
...
       "commit", "-m", shQuote("Initial commit")))
   })
-  module <- file.path("submodule", ".gitmodules")
-  on.exit(unlink(module), add = TRUE)
+
+  sub <- tempfile()
+  dir.create(sub)
+  on.exit(unlink(sub,recursive=TRUE,force=TRUE),add=TRUE)
+
+  module <- file.path(sub, ".gitmodules")
```
gaborcsardi, line 101: Are these changes necessary?

tests/testthat/test-submodule.R
```diff
@@ -93,43 +93,47 @@ test_that("Can install a repo with a submodule", {
...
       URLencode(dir),
       URLencode(dir)
     )
   )
 
   # The bar submodule is in .Rbuildignore, so we will not fetch it
-  build_ignore <- file.path("submodule", ".Rbuildignore")
-  on.exit(unlink(build_ignore), add = TRUE)
-
+  build_ignore <- file.path(sub, ".Rbuildignore")
```
gaborcsardi, line 117: Is this needed? You at least should keep the `on.exit()`, otherwise temporary files are not cleaned up.

tests/testthat/test-submodule.R
```diff
@@ -93,43 +93,47 @@ test_that("Can install a repo with a submodule", {
...
+  unlink(file.path(sub, "R"), recursive = TRUE, force = TRUE)
 
   # Install the package to a temporary library and verify it works
   lib <- tempfile()
   on.exit(unlink(lib, recursive = TRUE, force = TRUE), add = TRUE)
   dir.create(lib)
 
-  install_local("submodule", lib = lib, quiet = TRUE)
+  DESC_file <- file.path(sub,"DESCRIPTION")
+  writeLines("Package: submodule\nVersion: 0.0.0.9000",DESC_file)
```
gaborcsardi, line 134: Why is this now needed?

tests/testthat/test-submodule.R
```diff
@@ -143,19 +147,21 @@ test_that("Can update a submodule with an empty .gitmodules submodule", {
   dir.create(dir)
   on.exit(unlink(dir, recursive = TRUE, force = TRUE))
 
-  module <- file.path("submodule", ".gitmodules")
-  on.exit(unlink(module), add = TRUE)
+  sub <- tempfile()
+  dir.create(sub)
+  on.exit(unlink(sub,recursive=TRUE,force=TRUE),add=TRUE)
```
gaborcsardi, line 152: Can you please put spaces after the commas?

vignettes/dependencies.Rmd
```diff
@@ -41,7 +41,7 @@ Because GitHub is the most commonly used unofficial package distribution in R, i
 Remotes: r-lib/testthat
 ```
 
-You can also specify a specific hash, tag, or pull request (using the same syntax as `install_github()` if you want a particular commit. Otherwise the latest commit on the master branch is used.
+You can also specify a specific hash, tag, or pull request (using the same syntax as `install_github()` if you want a particular commit. Otherwise the latest commit on the default branch is used via a call to the HEAD commit.
```
gaborcsardi, line 44:
```suggestion
You can also specify a specific hash, tag, or pull request (using the same syntax as `install_github()` if you want a particular commit. Otherwise the latest commit on the default branch is used.
```

R/install-github.R
jimhester: Would it be more correct to say the default branch on the git remote? `HEAD` is a git construct, not really GitHub specific after all...

gaborcsardi: Yeah, sure. I would still say something like "default branch on GitHub" or similar, for the people that OK with GitHub but less familiar with git's internals, and maybe link to https://help.github.com/en/github/administering-a-repository/setting-the-default-branch ?

MyKo101: What about:
```
Defaults to `"HEAD"`, which means the default branch on GitHub/git. 
See [here](https://help.github.com/en/github/administering-a-repository/setting-the-default-branch) for more details.
```

tests/testthat/test-json.R
MyKo101: This is just testing that the JSON parser works and the test is passing.

tests/testthat/test-submodule.R
MyKo101: The test wasn't passing locally without these changes. `writeLines()` couldn't create the directory, so I used tempfile() to do so.

tests/testthat/test-submodule.R
MyKo101: The temporary files contained in `sub` would also be deleted when `sub` is deleted in the previous `on.exit()`

tests/testthat/test-submodule.R
MyKo101: The tests were failing without this. `install_local()` uses the `DESCRIPTION` file to get the package name. Without a `DESCRIPTION` file, `install_local()` wasn't working. Version also seemed to be needed as well as the package name.

tests/testthat/test-json.R
MyKo101: See previous comment

R/install-github.R
gaborcsardi: Well, `HEAD` only means the default branch for git remotes, not in git in general. E.g. if you have a git tree, `HEAD` means the tip of the current branch, not the default one. So:

... which means the default branch on GitHub and for git remotes.

MyKo101:
I've addressed the comments by @gaborcsardi and @jimhester in my local clone of `remotes`, but wanted to fix a few other things before pushing it.

I currently have two errors being hit by the `testthat.R` section of `devtools::check()`:
* `base curl download redirects (@test-download.R#298) `
  I believe that the url used here: `http://httpbin.org/absolute-redirect/1` is broken. Suggestions to replace would be useful.
* `install-github.R script is up to date (@test-script.R#10)`
  I'm unfamiliar with `make` and `brew`, so I'm finding it difficult to repair this. I have tried running `make` in terminal and `brew()` in the same way as it is run in the test, but it is still failing. It's also worth mentioning that when I run `devtools::test()`, it is passing. And in my previous commit, this worked (both locally and with appveyor/travis). It would be useful if someone could point me in the right direction (I am eager to learn, however I would also like a quick-fix for this and I feel like it is probably a really simple solution that I don't know how to implement)

There is also a warning being flagged when I run `devtools::test()`, but I'm not sure how important this is:
```
test-install-github.R:208: warning: github_pull
'tar.exe -xf "C:\Users\mbrxsmbc\AppData\Local\Temp\RtmpMFii7R\file3a587a373233.tar.gz" -C 
"C:/Users/mbrxsmbc/AppData/Local/Temp/RtmpMFii7R/remotes3a5854ae6762"' returned error code 1
```
  I have removed a reference to `"master"` in this test, however this `warning` remains with or without `"master"` included.

MyKo101:
Currently the only reason for the build failures is due to an issue with postmanlabs/httpbin#617.

jimhester:
Thanks a bunch!

PR #519, opened 2020-07-09
Add system_requirements function

jimhester:
This queries the public RStudio Package Manager for the system
requirements for a package


jimhester:
@gaborcsardi If you have any thoughts on other things that would be useful for this let me know. I think this is enough for us to use it on GitHub Actions at least.

Issue #530, opened 2020-08-17
`dev_package_deps` keeps only first `url_remote` dependency


dgabris:
We are using Github Enterprise in our company to host packages and their dependencies. We have 'package A' and in its DESCRIPTION file in repository we have specified the following dependencies:

```
Imports:
    B,
    C,
    D
Remotes:
    url::https://<company_github>.com/<user>/<B>/archive/master.zip,
    url::https://<company_github>.com/<user>/<C>/archive/master.zip,
    url::https://<company_github>.com/<user>/<D>/archive/master.zip
```

Then we install our package A using `remotes::install_github("<user>/<A>", host = "<company_github>/api/v3>")`.

This had some issues in the past (https://github.com/r-lib/remotes/issues/246), was fixed and working for some time, but now only the first url remote dependency is installed, while the other two are removed and considered to be on CRAN. This is happening on the last line of `dev_pockage_deps` function, [here](https://github.com/r-lib/remotes/blob/7b63f78c3fb2830d6feba5c73d5017bcf19de23b/R/deps.R#L147). The deduplication occurs, since the `url_remote` packages have `NA` package name and seem like the same dependency.

This is how `res` looks like before the last line of code:

![image](https://user-images.githubusercontent.com/33955390/90384104-56ce8700-e081-11ea-986a-4e8175a6b345.png)

And what is being returned from the function:

![image](https://user-images.githubusercontent.com/33955390/90384301-a1e89a00-e081-11ea-85f3-facc70707dbe.png)

Is this a bug, or should we be setting up our dependencies differently?

Thanks.

wwd1015:
Had similar issues but closer to the original issue #246. And proposed fix didn't work. Using the same example above, when you don't have remote package C or D installed first and try to install A directly, you will get the error message saying the package C or D is missing. It seem like the issue is that remotes will treat uninstalled package as CRAN package by default and therefore remote installation will not be triggered. This is the same as @dgabris mentioned above but the root cause is not in the ```dev_package_deps``` function. I was able to locate the issue but I'm not very familiar with the structure of this package. So hope someone can help fix this issue. 
https://github.com/r-lib/remotes/blob/e7630b8bc8ea4f2df39e8938b9fa127df4eab501/R/install-remote.R#L208-L215

PR #531, opened 2020-08-17
Remotes without package names are now unconditionally installed again

jakubkovac:
Fixes issue(#530).


jimhester:
Could you add a test so that we avoid a regression in this behavior again in the future?

jakubkovac:
I've added a simple a test. Unfortunately I wasn't able to save it without modifying the whitespace on line 183.

PR #538, opened 2020-09-06
Allow for GITHUB_TOKEN to be detected in the environment configuration

coatless:
Modifies `github_pat()` by allowing for a list of aliases if `GITHUB_PAT` isn't detected. Most notably, checks for the 
presence of `GITHUB_TOKEN` in the environment.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 * `install_*()` functions will no longer fail by default if there warnings from `install.packages()`. Concretely the default value of `R_REMOTES_NO_ERRORS_FROM_WARNINGS` has changed to `true` from the previous value of `false`. (#403)
 
+
+* `github_pat()` will now check if `GITHUB_TOKEN` is set if it cannot find `GITHUB_PAT`. (@coatless)
+
 # remotes 2.2.0
 
 ##  New functions and features
--- a/R/github.R
+++ b/R/github.R
@@ -64,18 +64,25 @@ github_commit <- function(username, repo, ref = "HEAD",
 #' Retrieve Github personal access token.
 #'
 #' A github personal access token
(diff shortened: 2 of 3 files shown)


coatless:
The error being detected by the test suite is independent of this change based on the CI failing for other PRs with the same error:

> ── 1. Failure: install-github.R script is up to date (@test-script.R#15)  ──────
readLines(tmp) not equal to readLines(file.path(root, "install-github.R")).
Lengths differ: 5560 is not 5553

jimhester:
Thanks for the PR!

In our workflow examples we set the `GITHUB_TOKEN` secret to the `GITHUB_PAT` environment variable [explicitly](https://github.com/r-lib/actions/blob/4fef3e012157302ad5012d6de16b64ac75d69386/examples/check-full.yaml#L40)

I think this is a better approach as even if we accepted this PR the `GITHUB_TOKEN` is a secret, not an environment variable. So you would still have to have a similar line in your workflow to set the `GITHUB_TOKEN` environment variable to enable it.

coatless:
@jimhester thanks for the quick review. The change here isn't necessarily for the CI routine but is due to bringing `remotes` recognization up-to-par with `gh`'s recognized variables. In particular, `gh` provided support for both `GITHUB_TOKEN` and `GITHUB_PAT`: https://github.com/r-lib/gh/blob/74655c4b171cf9e5512278759b6b408619f904aa/R/gh_token.R#L257-L263

That said,`GITHUB_TOKEN` appears in some [428k+ code files on GitHub](https://github.com/search?q=GITHUB_TOKEN&type=Code) in comparison to `GITHUB_PAT`'s only [10K that uses](https://github.com/search?q=GITHUB_PAT&type=Code).

jimhester:
Ok, makes sense, thanks again

Issue #448, opened 2019-10-17
Manage dependencies from GitHub Enterprise


hector-ps:
Dependencies from GitHub can be managed by including `Remotes: [remote::]repo_spec` in the `DESCRIPTION` file. Downloading packages from GitHub Enterprise can be downloaded requires the argument `host` on `install_github`, that cannot be defined in the `DESCRIPTION` file. While I understand that the general approach for packaging and sharing R code is using public repositories, company policies usually forces the developer to use private GitHub Enterprise. Are you considering to add a way for indicating inside the `DESCRIPTION` file the `host` to use?

cderv:
@ijlyttle How wonder how you do that with `remotes` and your 📦  `ghentr` ? 
Is this possible for you ?

ijlyttle:
The package maintainers can speak on this with infinitely-greater authority than I can: I think that extensibility mechanisms are something they are thinking about across r-lib, e.g. `usethis`, `remotes`, etc. 

To address @cderv's point, the I can't get the `ghentr` idea to work with remotes, which let me to ask about a way forward and workarounds. I can summarize the workaround they have suggested to me, which work for me while the larger issues are worked out. 

The central idea is to use a `url::` remote. For example, the standard GitHub remote:

```
r-lib/usethis
```

Can be expressed, roughly equivalently, as:

```
url::https://github.com/r-lib/usethis/archive/master.zip
```

Using a URL means that I don't have access to specific branches or releases, but I can point it at a repository on GitHub Enterprise.


Hope this helps!

cderv:
> The central idea is to use a url:: remote. For example, the standard GitHub remote:

This is what I thought. I got the same idea. I use that for gitlab on my side. Not the best way. 

It would be nice to have an option to tell remotes : « now we mean for github this url not github.com ». I guess it means a way to change globally the default host used in ˋinstall_github` and implicitly in ˋgithub_remoteˋ when parsing DESCRIPTION. 

Thanks @ijlyttle for you thoughts.

hector-ps:
Thanks @ijlyttle and @cderv for this workaround. I will use the `url::` remote until an _option to tell remotes : « now we mean for github this url not github.com »._ is developed. This is a really useful approach.

jimhester:
Just a small clarification, you can access releases and branches, e.g.

A specific release

https://github.com/r-lib/remotes/archive/v2.1.0.zip

A branch

https://github.com/r-lib/remotes/archive/always-install-uninstalled.zip

[...]

PR #541, opened 2020-09-21
Respect custom hosts

dagola:
In #448  a feature to specify a custom host URL in the Remotes field for e.g. enterprise GitHub or GitLab instances is requested.
This pull request adds a feature to specify the host URL explicitly:
```
<type>[@host]::<username>/<repo>[@ref]
```

The tests added may be not complete in the sense that I was not able to test against a custom host. Maybe anyone can provide some hints for better tests?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # remotes (development version)
 
+* It is now possible to specify a custom host for dependencies listed in the `DESCRIPTION` file with `Remotes: <type>[@host]::<username>/<repo>[@ref]`. The `ref` now supports `/` in it for `GitLab` repositories as it did for `GitHub`repositories. (@dagola, #448)
+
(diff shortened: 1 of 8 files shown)


dagola:
I also added a fix such that `ref`s with a slash are also allowed.

dagola:
@jimhester  Any problems with this MR? Can I improve this MR somehow so it gets merged?

jimhester:

dagola:
@jimhester I added a bullet to NEWS but the automated checks fail because of the test `internal map is current` in `test-bioc.R` fails. Looks like there is a new `bioconductor` version.

dagola:
@jimhester Anything else I can do?

jimhester:
Thanks a watt!

fabiandistlerkb:
That was really helpful, since I could not find it in the docs. Would you mind adding it, or point me to the place where this is documented? Especially the @host.name part was what I needed on ghe to make this work.

PR #544, opened 2020-10-07
Add more info on `upgrade` argument

Bisaloo:
This PR addresses the fact that the current documentation is a bit unclear when checking [`?install_github`](https://remotes.r-lib.org/reference/install_github.html). It is self-explanatory in the context of `package_deps()` but not so much in functions that inherit from it.

Please let me know if you'd like me to rewrap the source file and roxygenise. I didn't do it to keep my changes clear.


jimhester:
Thanks a bunch!

PR #549, opened 2020-10-28
Support name + version as os argument

krlmlr:
to `system_requirements()` . This allows using `os = "${{ matrix.config.os }}"` in GitHub Actions.


jimhester:
This would be simpler if implemented using `strsplit()`, or if you really wanted to use regular expressions `regexpr()`, so you would only have to run the regular expression once rather than three times.

PR #559, opened 2020-12-03
add space after colon before user input

daattali:
IMO it feels better as a user to have a space there

### DIFF
--- a/install-github.R
+++ b/install-github.R
@@ -1187,7 +1187,7 @@ function(...) {
     )
   }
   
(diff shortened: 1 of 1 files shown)


jimhester:
This isn't the right file for this, you need to change https://github.com/r-lib/remotes/blob/e7630b8bc8ea4f2df39e8938b9fa127df4eab501/R/deps.R#L658, then run `make` to update this file.

jimhester:
And also this is a duplicate of https://github.com/r-lib/remotes/pull/555.

daattali:
Thanks Jim! Yeah that's exactly what I was thinking too thanks @lionel-

Issue #578, opened 2021-02-03
R_BIOC_VERSION not respected in bioc_git2r_remote()?


bbimber:
I have an R package with testing via github actions. I would like to have one test running on bioconductor release and one on devel. My workflow does a number of things:

- sets "options(repos = BiocManager::repositories()"
- set Sys.setenv(R_BIOC_VERSION=BiocManager::version());

and then my DESCRIPTION lists the Bioconductor repos without specifying a branch (since this would make it hard to test across bioc releases), such as:

Remotes:
    bioc::DropletUtils

This is extremely close to working as I want; however, i figured out that when remote downloads package source from the bioconductor git, it's still using release. [Such as here](https://github.com/r-lib/remotes/blob/665ac7f3e0e55200e8b9b62992d8abaa2409be60/R/install-bioc.R#L117), [or here](https://github.com/r-lib/remotes/blob/665ac7f3e0e55200e8b9b62992d8abaa2409be60/R/install-bioc.R#L257). 

Is there another hook to change that? Otherwise, would it make sense for the latter to support something like:
```
    if (is.null(release)) {
      release <- getOption("Bioc_git_branch", "release")
    }
    if (release == "release") {
      release <- bioconductor_release()
    }
```
This requires two settings (the environment variable and options()), but I didnt immediately see how to effectively translate the numeric version number (which is what R_BIOC_VERSION provides) to devel/HEAD when that version is develop.

Is there some existing solution I'm missing? Thanks in advance.

bbimber:
FWIW, if some hook is enabled, some code like this would work in a github workflow. 

BiocManager::install(version = "${{ matrix.config.bioc }}", ask = FALSE);
cat(append = TRUE, file = "~/.Rprofile", "options(repos = BiocManager::repositories());")
cat(append = TRUE, file = "~/.Rprofile", "Sys.setenv(R_BIOC_VERSION=BiocManager::version());");
if (BiocManager::version() == BiocManager:::.version_bioc('devel')) {cat(append = TRUE, file = "~/.Rprofile", "options(Bioc_git_branch = 'master');") }

jimhester:
You should set a `BiocViews: ` field in your `DESCRIPTION`, and then the Bioconductor packages will be automatically installed from the Bioconductor repositories returned by `BiocManager`. You generally don't need to use an explicit bioc remote unless you are developing multiple Bioconductor packages in tandem.

bbimber:
[...]

PR #580, opened 2021-02-03
Switch branch detection to respect R_BIOC_VERSION

bbimber:
@jimhester Thanks for replying on the issue. My original idea here involved making a redundant option() to hold the git branch used when querying bioconductor. This updated PR simply uses the already supported R_BIOC_VERSION environment variable, and ensures that when querying git we use a branch that respects that setting.

There seem to be tests that need updating, which i would do if this is something you'd consider. 

The remotes/bioconductor configuration is complex and maybe I'm missing something. However, if R_BIOC_VERSION is supposed to force remotes to use a specific version it seems like it should be respected on this codepath too.

### DIFF
--- a/R/install-bioc.R
+++ b/R/install-bioc.R
@@ -254,7 +254,7 @@ bioconductor_branch <- function(release, sha) {
     sha
   } else {
     if (is.null(release)) {
-      release <- ifelse(Sys.getenv("R_BIOC_VERSION") != "", yes = Sys.getenv("R_BIOC_VERSION"), no = "release")
+      release <- Sys.getenv("R_BIOC_VERSION", "release")
     }
     if (release == "release") {
       release <- bioconductor_release()
--- a/R/install-bioc.R
+++ b/R/install-bioc.R
@@ -259,7 +259,7 @@ bioconductor_branch <- function(release, sha) {
     if (release == "release") {
       release <- bioconductor_release()
     } else if (release == bioconductor$get_devel_version()) {
-      release = "devel"
+      release <- "devel"
     }
     switch(
       tolower(release),
@@ -301,4 +301,3 @@ git_repo_sha1 <- function(r) {
     git2r::branch_target(rev)
   }
 }
-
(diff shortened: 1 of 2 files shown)


bbimber:
@jimhester once again, thanks for your replies on #578. I wanted to follow up here - if the intention of the environment variable R_BIOC_VERSION is to force a specific bioconductor version, I believe it should fully respect that setting in bioconductor_branch() too, which this PR adds. Unless I'm missing something, currently even if R_BIOC_VERSION is set, bioconductor_branch() is always going to pull from the git release branch. This is a very discrete update that defers to R_BIOC_VERSION if set.

R/install-bioc.R
```diff
@@ -254,10 +254,12 @@ bioconductor_branch <- function(release, sha) {
     sha
   } else {
     if (is.null(release)) {
-      release <- "release"
+      release <- ifelse(Sys.getenv("R_BIOC_VERSION") != "", yes = Sys.getenv("R_BIOC_VERSION"), no = "release")
```
jimhester, line 257:
```suggestion
      release <- Sys.getenv("R_BIOC_VERSION", "release")
```

bbimber: good point - changed

R/install-bioc.R
```diff
@@ -254,10 +254,12 @@ bioconductor_branch <- function(release, sha) {
     sha
   } else {
     if (is.null(release)) {
-      release <- "release"
+      release <- Sys.getenv("R_BIOC_VERSION", "release")
     }
     if (release == "release") {
       release <- bioconductor_release()
+    } else if (release == bioconductor$get_devel_version()) {
+      release = "devel"
```
jimhester, line 262:
```suggestion
      release <- "devel"
```

jimhester: This matches the style used in remotes.

jimhester:
Thanks, can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

bbimber:
@jimhester OK, added NEWS

jimhester:
Thank you!

PR #608, opened 2021-05-01
Adding `git_remote` fallback for `gitlab_remote` use without full API access (Resolves #604)

### LINKED ISSUE #604: [Feature Request] Default to `install_git` when access token can't use Gitlab repository/ API endpoint
A bit of a goofy behavior, but Gitlab's `CI_JOB_TOKEN` (the access token created for the execution of each CI job), has very [selective access to the Gitlab API](https://docs.gitlab.com/ee/api/README.html#gitlab-cicd-job-token). 

### Errors using `install_gitlab` in private gitlab CI

Notably, it does not have access to the [`projects/:id/repository/files` endpoint](https://docs.gitlab.com/ee/api/repository_files.html) or [`projects/:id/repository/archive` endpoint](https://docs.gitlab.com/ee/api/repositories.html#get-file-archive) used to fetch the [remote `DESCRIPTION`](https://github.com/r-lib/remotes/blob/427719eb56860e0131f2ad9ac91389ea54625fcb/R/install-gitlab.R#L125-L134) and download the [package archive](https://github.com/r-lib/remotes/blob/427719eb56860e0131f2ad9ac91389ea54625fcb/R/install-gitlab.R#L82-L90). These are accessible on public Gitlab because these endpoints are accessible without authentication. However, any private instance which requires authentication will fail using this token. Users would have to embed personal tokens with a CI job in order to use `install_gitlab` - feasible, but a poor practice as personal tokens would be exposed to co-maintainers/owners. 

Oddly enough, the `CI_JOB_TOKEN` _does_ have full ability to `git clone` any unrestricted private repo using basic auth (`git clone https://gitlab-ci-token:$CI_JOB_TOKEN@example.com/<namespace>/<project>.git`). This can be used in any CI job after setting

```r
options(remotes.git_credentials = git2r::cred_user_pass("gitlab-ci-token", Sys.getenv("CI_JOB_TOKEN"))`)
```

### Suggested fallback

When the [`projects/:id/repository/files`](https://docs.gitlab.com/ee/api/repositories.html#get-file-archive) endpoint returns a `401` response, it would be wonderful if `remotes` would fall back to using `install_git` in its place. This would allow use of `install_gitlab` within CI jobs without having to embed personal access tokens within the repo's env vars. 

### PR Considerations

I'd be happy to help put together a PR, but I would appreciate some feedback about approach. As is, the first failed request isn't hit until `remote_package_name`, but I don't think that's the right point to handle the fallback. Alternatively, this endpoint could get hit even before returning a `remote` object from `install_gitlab`, allowing an easy point to fallback, but possibly duplicating a request. 

If you'd be interested in accepting this functionality, I would greatly appreciate some thoughts on the design before starting to put something together.

dgkf:
As described in #604, the current `gitlab_remote` makes use of API endpoints that are not available to tokens generated for use within gitlab CI (stored in the [`$CI_JOB_TOKEN`](https://docs.gitlab.com/ee/api/README.html#gitlab-cicd-job-token) env var), throwing errors when these tokens are used.

This PR adds code to first ping the API at a generic endpoint (querying for [/version](https://docs.gitlab.com/ee/api/version.html)). If that request fails and `isTRUE(getOption("remotes.gitlab_git_fallback", TRUE))`, a `git_remote` is returned.

If `git2r` is available, a credentials object is created from the `auth_token`. Otherwise, the token is embedded in the url in the form of `http://gitlab-ci-token:$TOKEN@example.com/namespace/project.git`.

This allows `install_gitlab` to be used within CI jobs on non-public deployments of GitLab without the creation and embedding of personal tokens. Pipeline engineers need only to run `export GITLAB_PAT=$CI_JOB_TOKEN` prior to installing remotes.

### DIFF
--- a/R/install-gitlab.R
+++ b/R/install-gitlab.R
@@ -63,7 +63,7 @@ gitlab_remote <- function(repo, subdir = NULL,
   meta <- parse_git_repo(repo)
   meta$ref <- meta$ref %||% "HEAD"
 
-  if (auth_token_has_gitlab_api_access(host = host, pat = auth_token)) {
+  if (auth_token_has_gitlab_api_access(host = host, pat = auth_token) || !isTRUE(git_fallback)) {
     remote("gitlab",
       host = host,
       repo = paste(c(meta$repo, meta$subdir), collapse = "/"),
@@ -73,7 +73,7 @@ gitlab_remote <- function(repo, subdir = NULL,
       sha = sha,
       auth_token = auth_token
     )
-  } else if (isTRUE(git_fallback)) {
+  } else {
     credentials <- git_credentials()
     url <- paste0(build_url(host, repo), ".git")
 
--- a/R/install-gitlab.R
+++ b/R/install-gitlab.R
@@ -57,20 +57,55 @@ install_gitlab <- function(repo,
 
 gitlab_remote <- function(repo, subdir = NULL,
                        auth_token = gitlab_pat(), sha = NULL,
-                       host = "gitlab.com", ...) {
+                       host = "gitlab.com", ..., 
+                       git_fallback = getOption("remotes.gitlab_git_fallback", TRUE)) {
 
   meta <- parse_git_repo(repo)
   meta$ref <- meta$ref %||% "HEAD"
 
-  remote("gitlab",
-    host = host,
-    repo = paste(c(meta$repo, meta$subdir), collapse = "/"),
-    subdir = subdir,
-    username = meta$username,
-    ref = meta$ref,
-    sha = sha,
-    auth_token = auth_token
-  )
+  if (auth_token_has_gitlab_api_access(host = host, pat = auth_token)) {
+    remote("gitlab",
+      host = host,
+      repo = paste(c(meta$repo, meta$subdir), collapse = "/"),
+      subdir = subdir,
+      username = meta$username,
+      ref = meta$ref,
+      sha = sha,
+      auth_token = auth_token
+    )
+  } else if (isTRUE(git_fallback)) {
+    credentials <- git_credentials()
+    url <- paste0(build_url(host, repo), ".git")
+
+    git2r_inst <- pkg_installed("git2r")
+    has_token  <- !is.null(auth_token)
+    url_has_token <- grepl("^(.*://)?[^@/]+@", url)
+
+    if (git2r_inst && has_token) {
+      credentials <- getExportedValue("git2r", "cred_user_pass")(
+        username = "gitlab-ci-token",
+        password = auth_token
+      )
+    } else if (!url_has_token && !git2r_inst && has_token) {
+      url_protocol <- gsub("((.*)://)?.*", "\\1", url)
+      url_path     <- gsub("((.*)://)?", "", url)
+      url <- paste0(
+        url_protocol,
+        "gitlab-ci-token:",
+        auth_token,
+        "@",
+        url_path
+      )
+    }
+
+    git_remote(
+      url = url,
+      subdir = subdir,
+      credentials = credentials,
+      ref = sha %||% meta$ref,
+      ...
+    )
+  }
 }
 
 #' @export
@@ -158,6 +193,19 @@ gitlab_commit <- function(username, repo, ref = "HEAD",
   download(tmp, url, headers = c("Private-Token" = pat))
 
   json$parse_file(tmp)$id
+}
+
+auth_token_has_gitlab_api_access <- function(host = "gitlab.com", pat) {
+  # use the /version endpoint - general access endpoint with small payload, but 
+  # inaccessible to CI tokens
+  url <- build_url(host, "api", "v4", "version")
+  has_access <- tryCatch({
+    download(tempfile(), url, headers = c("Private-Token" = pat))
+    TRUE
+  }, error = function(e) {
+    FALSE
+  })
+  has_access
 }
 
 #' Retrieve GitLab personal access token.


jimhester:
My main worry with making the GitLab remote more complex is our team doesn't use GitLab, so it is possible this will break in the future without us realizing it.

dgkf:
Thanks @jimhester - I'm happy to add in tests as much as possible. If you think the implementation looks sound, then I can get to work on tests and updating docs. I was just hesitant to invest more time fleshing out the peripheral bits until getting some impressions on the approach.

dgkf:
@jimhester - this PR is ready whenever you have an opportunity to take a look. Overall, the design feels a bit clunky, but I'm struggling to come up with anything better. 

To trace through the changes, it is easiest to start with `install_gitlab`'s call out to `gitlab_to_git_remote`, and then look at the uses of `$url` in `install_git.R` as the url may contain a url-embedded username and token.

Just to highlight a critical design choice:

**Design Feedback Request**: is it preferred to keep the full url including username and password (`https://dgkf:12345@gitlab.com/...`) when storing `remote_metadata` or printing to console? 
- _the good_: this would allow `remotes::update_packages` to update a package which requires authentication
- _the bad_: this might put a user at risk of leaking an access token through things like execution logs if printed to console or an renv lockfile if included in a DESCRIPTION file.  

For now, I chose to scrub the username and password from the url before this is added to a DESCRIPTION file to prioritize safety of access tokens over the update experience. 

---

The self-hosted GitLab issues are currently a big pain point at my org, so some help in moving this forward would be greatly appreciated.

statnmap:
Thank you for this PR. This is a must when working on private GitLab instances.
I approve its improvements.

I tried it in a CI instance with the following classical use cases I guess. The PR solves the problems encountered with current version of {remotes}. This can be accepted as is.

# Use `CI_JOB_TOKEN` set up with {git2r}

## Clone and `install_local()`

- current {remotes} OK
- PR OK
```r
tempclone <- tempfile(pattern = "conjdown")
dir.create(tempclone)

git2r::clone(url = "https://git.lab.sspcloud.fr/propre-conj/conjdown",
             local_path = tempclone,
             credentials = git2r::cred_user_pass(username = "gitlab-ci-token",
                                                 password = Sys.getenv("CI_JOB_TOKEN"))
)

remotes::install_local(tempclone)
```

## install_git()
- current {remotes} FAIL
- PR OK
```r
options(remotes.git_credentials = git2r::cred_user_pass("gitlab-ci-token", Sys.getenv("CI_JOB_TOKEN")))
  remotes::install_git("https://myprivategitlab.com/user/repos")
```

## install_gitlab()

- current {remotes} FAIL
- PR OK with message
```r
  remotes::install_gitlab(host = "https://myprivategitlab.com",
                          repo = "user/repos",
                          auth_token = Sys.getenv("CI_JOB_TOKEN"))
```
message:
```
auth_token does not have scopes 'read-repository' and 'api' for host
  'https://myprivategitlab.com" required to install using
  gitlab_remote.
Attempting git_remote
```

## install from another package DESCRIPTION file with git2r creds and `git::`

- current {remotes} FAIL
- PR OK 

*DESCRIPTION file*
```
Imports: 
    repos
Remotes:
    git::https://myprivategitlab.com/user/repos"
```
```r
options(remotes.git_credentials = git2r::cred_user_pass("gitlab-ci-token", Sys.getenv("CI_JOB_TOKEN")))
remotes::install_deps(dependencies = TRUE)
```

# Set `GITLAB_PAT`
## install_gitlab()

- current {remotes} FAIL
- PR OK with message
```r
Sys.setenv(GITLAB_PAT = Sys.getenv("CI_JOB_TOKEN"))
  remotes::install_gitlab(host = "https://myprivategitlab.com",
                          repo = "user/repos")
```
message:
```
Using GitLab PAT from envvar GITLAB_PAT
auth_token does not have scopes 'read-repository' and 'api' for host
  'https://myprivategitlab.com" required to install using
  gitlab_remote.
Attempting git_remote
```

## install from another package DESCRIPTION file with GITLAB_PAT and `gitlab::`

This is a try. I know this is not the aim of this PR, but that could be a future enhancement, maybe.

- current {remotes} FAIL
- PR FAIL.

*DESCRIPTION file*
```
Imports: 
    repos
Remotes:
    gitlab::https://myprivategitlab.com/user/repos"
```
```r
Sys.setenv(GITLAB_PAT = Sys.getenv("CI_JOB_TOKEN"))
remotes::install_deps(dependencies = TRUE)
```
*Error*
```
Error: Unknown remote type: gitlab
  Invalid git repo specification: 'https://git.lab.sspcloud.fr/propre-conj/conjdown'
Execution halted
```

statnmap:
Do you think that it could be a good idea to allow `gitlab_pat()` to also look for `CI_JOB_TOKEN` environment variable if `GITLAB_PAT` is empty ? This may solve a lot pain using CI.

dgkf:
Thanks for considering this PR, @jimhester.

Just wanted to highlight this bit in the PR thread for your consideration. I tried my best to dig into how `remotes`/`renv` use the `Remotes*` fields in the `DESCRIPTION` file, but wasn't totally sure what the preferred solution would be for access tokens in urls and want to make sure it was brought to your attention in case there are any security concerns with how it's handled.

> **Design Feedback Request**: is it preferred to keep the full url including username and password (`https://dgkf:12345@gitlab.com/...`) when storing `remote_metadata` or printing to console? 
> - _the good_: this would allow `remotes::update_packages` to update a package which requires authentication
> - _the bad_: this might put a user at risk of leaking an access token through things like execution logs if printed to console or an renv lockfile if included in a DESCRIPTION file

Currently user-specific url components are stripped to minimize any printing/saving of tokens.

jimhester:
I guess we should strip them, though it would then break updating packages later.

However if you still set the `GITLAB_PAT` when you run `update_packages()` would the update work?

PR #609, opened 2021-05-02
Fix split of os and os-release in system_requirements, and add unit test

mdneuzerling:
`system_requirements` supports specifying an operating system like "ubuntu-16.04", in which case it's intended that the input is split along the hyphen and `os = "ubuntu", os_release = "16.04"`. Currently, the following code causes a "non-character argument" error when this is attempted, as `strsplit` attempts to run on a `NULL` value:

```r
  if (is.null(os_release)) {
    os_release <- strsplit(os_release, "-", fixed = TRUE)[[1]]
    # ...
  }
```

This PR changes the above line to `os_release <- strsplit(os, "-", fixed = TRUE)[[1]]`, and adds a unit test to ensure that we can determine system dependencies for the `curl` package when the operating system is specified as "ubuntu-16.04".

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -19,3 +19,4 @@
 ^CRAN-RELEASE$
 ^\.gitattributes$
 ^\.github$
+^LICENSE\.md$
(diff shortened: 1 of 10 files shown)


jimhester:
Thanks a watt!

PR #623, opened 2021-06-07
Update usethis function for token generation

### LINKED ISSUE #590: github_error() refers to deprecated usethis function
When a user exceeds the GitHub API limit, an error message (helpfully) directs them to add a GitHub token to their R environment. However, the error message suggests that the user uses `usethis::browse_github_pat()`, which has been deprecated in favour of `usethis::create_github_token()`.

The relevant line:
https://github.com/r-lib/remotes/blob/665ac7f3e0e55200e8b9b62992d8abaa2409be60/R/github.R#L160

malcolmbarrett:
Closes #590


jimhester:
Thank you!

Issue #621, opened 2021-06-04
install_git() fails to install package from AzureDevOps with command status 128


mima25:
We´ve been using remotes v2.3 to install internally developed packages that are stored on AzureDevOps. Unable to use the function since the package upgrade with the following error:
`remotes::install_git( 'git@ssh.dev.azure.com:v3/dataapi-rclient', git = 'external')`
Error: Failed to install 'unknown package' from Git:
  Command failed (128)
In addition: Warning message:
In system(full, intern = TRUE, ignore.stderr = quiet) :
  running command ''/usr/bin/git' ls-remote git ssh.dev.azure.com:v3/dataapi-rclient 2>/dev/null' had status 128

`session info():`
R version 4.1.0 (2021-05-18)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.2 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.9.0
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.9.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] remotes_2.4.0

ellisp:
I have a similar problem with this code (unfortunately not a reproducible example) for installing from a private GitLab repository, which works under 2.3.0 but does not work under 2.4.0:

```R
uid <- getPass::getPass("Enter your GitLab username (not your email address but the actual GitLab name)", noblank = TRUE)

remotes::install_git(
  url = "https://gitlab.com/nous-group/ip/nous.utils",
  credentials = git2r::cred_user_pass(uid, getPass::getPass("GitLab password:")),
  build_opts = c("--no-resave-data", "--no-manual"),
  dependencies = TRUE,
  upgrade = TRUE
)
```

The error message is

```
Error: Failed to install 'unknown package' from Git:
[...]

PR #628, opened 2021-06-16
Adding error handling to git_remote http requests (Resolves #625)

### LINKED ISSUE #625: install_git() Assumes github-style urls and credential-less web server access
Despite being able to clone a repository in the command line, I was running into errors using `install_git` to install dependencies. After some investigation, I found a few assumptions that might contribute to errors:

https://github.com/r-lib/remotes/blob/a14cf0031da2d083854246914b4974f09656443a/R/install-git.R#L148-L151

These lines in `remote_package_name.git2r_remote` assume a github-style raw file content url without a safe fallback. ~~For example, where GitHub will use `.../raw/<ref>/<path>`, GitLab will use `.../-/raw/<ref>/<path>`.~~ _Edit: after further testing, the GitHub style is totally fine for GitLab._

Likewise, this assumes that one has unauthenticated access to these web pages. At least in my org, I run into the error below because `read.dcf` is trying to parse the log-in page that unauthenticated access redirects to.

```
Error: Failed to install 'unknown package' from Git:
  Line starting '<!DOCTYPE html> ...' is malformed!
```

Would it be possible to fall back to return `NA` or fall back to a method that only assumes a basic git functionality when such assumptions cause errors?

dgkf:
Addressing #625

- Simple error handling, returning `NA` if either the download or DESCRIPTION parsing fails. 
- Tests for a gitlab repo and an inaccessible repo (just given a fake url) to simulate an http endpoint behind authentication. (had to test to make sure the only warnings that get emitted are from `remote_sha` since it's not a real git url, but otherwise tests the http recovery). 

Happy to iterate from here, but felt it was simple enough to just get the ball rolling on it.

### DIFF
--- a/R/install-git.R
+++ b/R/install-git.R
@@ -152,7 +152,7 @@ remote_package_name.git2r_remote <- function(remote, ...) {
       download(tmp, url)
       read_dcf(tmp)$Package
     }, error = function(e) {
-      NA_character_      
+      NA_character_
     })
   } else {
     # Try using git archive --remote to retrieve the DESCRIPTION, if the protocol
--- a/R/install-git.R
+++ b/R/install-git.R
@@ -146,9 +146,14 @@ remote_package_name.git2r_remote <- function(remote, ...) {
   description_path <- paste0(collapse = "/", c(remote$subdir, "DESCRIPTION"))
 
   if (grepl("^https?://", remote$url)) {
+    # assumes GitHub-style "<repo>/raw/<ref>/<path>" url
     url <- build_url(sub("\\.git$", "", remote$url), "raw", remote_sha(remote, ...), description_path)
-    download(tmp, url)
-    read_dcf(tmp)$Package
+    tryCatch({
+      download(tmp, url)
+      read_dcf(tmp)$Package
+    }, error = function(e) {
+      NA_character_      
+    })
   } else {
     # Try using git archive --remote to retrieve the DESCRIPTION, if the protocol
     # or server doesn't support that return NA
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: remotes
 Title: R Package Installation from Remote Repositories, Including 'GitHub'
-Version: 2.4.0.9000
+Version: 2.4.0.9001
 Authors@R: c(
     person("Jim", "Hester", , "jim.hester@rstudio.com", role = c("aut", "cre")),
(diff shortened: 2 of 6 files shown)


R/install-git.R
```diff
@@ -146,9 +146,14 @@ remote_package_name.git2r_remote <- function(remote, ...) {
...
   if (grepl("^https?://", remote$url)) {
+    # assumes GitHub-style "<repo>/raw/<ref>/<path>" url
     url <- build_url(sub("\\.git$", "", remote$url), "raw", remote_sha(remote, ...), description_path)
-    download(tmp, url)
-    read_dcf(tmp)$Package
+    tryCatch({
+      download(tmp, url)
+      read_dcf(tmp)$Package
+    }, error = function(e) {
+      NA_character_
```
niheaven, line 155:
```suggestion
      return(NA_character_)
```

Yes I know that will just return 😄 But let's make it clear~

niheaven (approved):
LGTM 👍

R/install-git.R
jimhester: In general we don't put explicit returns except for early returns. See https://style.tidyverse.org/functions.html#return for details.

jimhester:
Thanks @dgkf for the PR and @niheaven for reviewing.

@dgkf if you could add a note to the NEWS I can merge this.

dgkf:
Should be good to go - I just added a note to @niheaven's description of the HTTP changes. I felt that an additional NEWS entry would be unclear since it only pertains to new features in the upcoming release. Let me know if you'd prefer that it's a separate entry and I can split it out.

niheaven:
@dgkf Maybe a separate entry is preferred, since my PR has already been merged into 2.4.0, and this PR sould be there in next version.

dgkf:
Thanks @niheaven - mistake on my part. Didn't realize that code was already in a release. I reverted my edits and broke out the NEWS entry into the dev section.

dgkf:
As far as I can tell, any errors on windows builds are hold overs from issues on current master. Let me know if there's anything more that you'd like me to address.

hokkwan:
I hope this PR can be incorporated soon, because this also resolves the issue regarding installing packages from Git repositories on AzureDevOps, see issue #621.

jimhester:
Hey! Thanks!

Thanks @dgkf for working on this PR and for @niheaven for reviewing!

Issue #20, opened 2016-01-14
Test cases for new functions (see missing coverage)


gaborcsardi:

gaborcsardi:
Should be done.

PR #640, opened 2021-08-13
Clarify `github_pull()` documentation

ms609:
It is unclear to me what format is expected of `pull`: should I specify `20` or `"#20"`?

I believe from looking through the package source that the expected format is `"20"` and have proposed updating examples accordingly.


jimhester:
Thanks!

PR #671, opened 2021-11-29
Use gitcreds for the GH PAT

### LINKED ISSUE #587: Use the git credential store
Cf. https://github.com/r-lib/devtools/issues/2299

gaborcsardi:
Closes #587.

We still check the env vars first, so the messages re the PAT source are better.

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -80,8 +80,9 @@ github_pat <- function(quiet = TRUE) {
     if (nzchar(pat)) {
       if (!quiet) {
         message("Using github PAT from envvar ", env_var, ". ",
-                "Use `gitcreds::gitcreds_set()` if you want to use the git ",
-                "credential store instead")
+                "Use `gitcreds::gitcreds_set()` and unset ", env_var,
+                " in .Renviron (or elsewhere) if you want to use the more ",
+                "secure git credential store instead.")
       }
       return(pat)
     }
--- a/R/github.R
+++ b/R/github.R
@@ -79,10 +79,23 @@ github_pat <- function(quiet = TRUE) {
     pat <- Sys.getenv(env_var)
     if (nzchar(pat)) {
       if (!quiet) {
-        message("Using github PAT from envvar ", env_var)
+        message("Using github PAT from envvar ", env_var, ". ",
+                "Use `gitcreds::gitcreds_set()` if you want to use the git ",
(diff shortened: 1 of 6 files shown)


R/github.R
```diff
@@ -79,20 +79,33 @@ github_pat <- function(quiet = TRUE) {
     pat <- Sys.getenv(env_var)
     if (nzchar(pat)) {
       if (!quiet) {
-        message("Using github PAT from envvar ", env_var)
+        message("Using github PAT from envvar ", env_var, ". ",
+                "Use `gitcreds::gitcreds_set()` if you want to use the git ",
+                "credential store instead")
```
jennybc, line 84: Elsewhere, I've also tried to nag people to proactively undo whatever was setting `GITHUB_PAT` or `GITHUB_TOKEN` in the first place, e.g. to open `.Renviron` and delete the relevant line. It feels like you should probably do same, if you're going to open this can of worms.

R/github.R
```diff
@@ -79,20 +79,33 @@ github_pat <- function(quiet = TRUE) {
...
       }
       return(pat)
     }
   }
 
+  pat <- tryCatch(
+    gitcreds_get()$password,
```
jennybc, line 91: It feels like the `host` should appear here, i.e. what if we're installing from a GitHub Enterprise deployment? I realize this implies that `github_pat()` needs to gain a `host` argument.

Maybe the git credential store is only supported / consulted when `host == "api.github.com"`?

I've verified that this PR can send my github.com PAT to a different host (and, obviously, the install fails).

gaborcsardi: Yes, that is the proper way, but that would need a major refactoring of remotes, which is probably not worth doing.

remotes already sends the token to wrong host, so not change of behavior there. The only change is that we also check the credential store now.

R/github.R
gaborcsardi: Yeah. I figured (some) people probably have a good reason to use the env var. (e.g. cron job on Linux, or just Linux in general, where the credential helpers are really lacking), so I did not want to annoy them too much with a longer message. But I'll make it a bit longer to note that they can delete the env var.

R/github.R
jennybc: Yes, everything you say makes sense.

PR #685, opened 2022-01-03
add option `remotes.download.headers`

### LINKED ISSUE #682: available_packages has no option to set `headers`
We use a private CRAN repository which requires http authorization headers to query and install packages.
While `remotes::install_deps()` has a `headers` argument, it is not passed to `remotes::available_packages()`, preventing correct operation.

A clumsy work around is manually setting the cache beforehand via

```
ap <- utils::available.packages(type = "source", headers = my_secret_auth)
remotes::available_packages_set(
  repos = getOption("repos"),
  type = "source",
  db = ap
)
remotes::install_deps(
  dependencies = TRUE,
  type = "source",
  headers = my_secret_auth
)
```

It would be nice if `remotes::install_deps()` passed the headers option to `remotes::available_packages()` so this work-around is no longer necessary.

AshesITR:
this allows passing e.g. `headers` to both, `utils::install.packages()` and `utils::available.packages()`

If you don't like passing ellipsis twice, I could instead make `headers` a named argument in all functions and pass that down to both functions while keeping ellipsis only for the call to `install.packages()`.
The downside to this is that the argument might be surprising given that no other networking-related arguments are exposed in the same way.

fixes #682


gaborcsardi:
Passing the `...` twice is not great because `available.packages()` and `install.packages()` take different arguments. Instead top level functions could have an `available_packages_args` argument, which is a list of arguments that we pass to `available.packages()`.

Alternatively, we could have an option to set the headers that are passed to `available.packages()` and/or `install.packages()`.

AshesITR:
I like the option alternative. It is similar to `renv`, which uses `renv.download.headers` [described here](https://rstudio.github.io/renv/articles/renv.html#authentication-with-custom-headers).
Will rework this.

Should I add the `headers` argument to top-level functions so e.g. `remotes::install_deps(headers = ...)` also works as expected?

gaborcsardi:
I think if we have an option, then we don't need to add new arguments. You'll need to work out what to do if the user already supplies the `headers` argument in `...`, though. Maybe merge them with the argument taking precedent?

AshesITR:
@gaborcsardi could you take another look? I've implemented it via the character option `remotes.download.headers`.
If `headers` was manually provided to any install command, a message is generated pointing to the feature (I'm afraid it's not trivially possible to capture the `headers` argument in all the right places without capturing it too often).
If we wanted to implement the function interface renv uses, we'd have to also re-implement `utils::available.packages` like renv does, so I didn't go that route.

Thanks!

Issue #676, opened 2021-12-03
Support syntax for remotes where repo name is different from package name


heavywatal:
As discussed in r-lib/pkgdepends#263, it would be nice if `remotes` could also supports a new format suggested in r-lib/pkgdepends#2:
```
Imports:
    igraphlite
Remotes:
    igraphlite=github::heavywatal/rigraphlite
```

PR #719, opened 2022-08-30
Allow explicit package names in Remotes

heavywatal:
This package somehow manages differences in repo names vs package names. Ignoring the leading "pkgname=" part is enough to let this package work.

Addresses #676

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 
 * `install_github()` now uses credentials from the git credential store,
   if `GITHUB_PAT` and `GITHUB_TOKEN` are not set.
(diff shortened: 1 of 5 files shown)


gaborcsardi:
Great idea! I think we could make the implementation a bit safer by matching only something like `^[.a-zA-Z0-9]+=` instead of just a plain `=` which might appear in some remote specifications.

gaborcsardi:
Great, thanks! Can you please also add a bullet point to `NEWS.md`, mentioning this PR (or the issue) and your GH username?

heavywatal:
Done.

gaborcsardi:
Thank you!

Issue #732, opened 2022-11-04
Make remotes sources precedence optional


maksymiuks:
Hi,

I'd like to suggest an enhancement to how remote sources are handled by the `combine_remote_deps` function https://github.com/r-lib/remotes/blob/main/R/deps.R#L150-L164. In particular, I think it would be beneficial to make this line:

https://github.com/r-lib/remotes/blob/main/R/deps.R#L161

conditional on some parameter or the env variable. The rationale is that sometimes as a developer, you want to check how the package behaves against CRAN sources, disregarding remotes. Right now it's hard to reproduce such a state using this function as remotes always take precedence over the CRAN forcing a developer to alter DESCRIPTION just for this one scenario (which could be tedious for instance in automated builds). Therefore I think keeping default behaviour as is but allowing to actually skip remotes if CRAN is present (or just don't include remote sources at all, will suffice as well) will make the lives of some developers easier. 

Let me know what you think about this suggestion, happy to make a PR if you think its fine

gaborcsardi:
Yes, a PR would be great.

maksymiuks:
@gaborcsardi PR ready for the review https://github.com/r-lib/remotes/pull/733

maksymiuks:
Feature merged, thanks for a quick response!

PR #733, opened 2022-11-04
Add remote_precedence parameter to dev_package_deps

maksymiuks:
As discussed in the https://github.com/r-lib/remotes/issues/732 it's a PR with suggested implementation.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,9 +4,8 @@
 * `install_github()` now uses credentials from the git credential store,
   if `GITHUB_PAT` and `GITHUB_TOKEN` are not set.
 * Remotes field accepts explicit package names: `<pkgname>=<type>::<username>/<repo>` (#719, @heavywatal).
-* Add `remote_precedence` parameter to `dev_package_deps()` and `combine_remote_deps()` that allows
-  user to choose whether remote sources should have a priority over cran sources. Parameter defaults
-  to TRUE for backward compatibility reasons. (#732, @maksymiuks)
+* Add `remote_precedence` parameter to `dev_package_deps()` which allows user to choose whether 
+  remote sources should have a priority over cran sources. (#732, @maksymiuks)
 
 # remotes 2.4.2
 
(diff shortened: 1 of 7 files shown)


gaborcsardi:
Thanks! Hmmm, does this really help people install a package with the CRAN dependencies? How would one do that?

NEWS.md
```diff
@@ -4,6 +4,9 @@
 * `install_github()` now uses credentials from the git credential store,
   if `GITHUB_PAT` and `GITHUB_TOKEN` are not set.
 * Remotes field accepts explicit package names: `<pkgname>=<type>::<username>/<repo>` (#719, @heavywatal).
+* Add `remote_precedence` parameter to `dev_package_deps()` and `combine_remote_deps()` that allows
```
gaborcsardi, line 7: `combine_remote_deps()` is not exported, we should not mention it here.

NEWS.md
```diff
@@ -4,6 +4,9 @@
...
+  user to choose whether remote sources should have a priority over cran sources. Parameter defaults
+  to TRUE for backward compatibility reasons. (#732, @maksymiuks)
```
gaborcsardi, line 9: It is not "just" for backward compatibility, but rather because that's the usual case. So I would just remove this last sentence.

gaborcsardi (changes requested):

gaborcsardi:
> it will be helpful in any case when for some reason we need to update our dependencies to the version available on CRAN, disregarding remotes if possible.

Well, except that if the non-CRAN version is already installed, then it will not "update" that to the CRAN version.

maksymiuks:
I agree word "update" wasn't really a fortunate one. You are right, if the developer stores everything in one library then yes it won't be updated, as probably remote is ahead of CRAN so it will be skipped by the design of this function. But if one has isolated libraries for testing and properly sets the search path, or if it's a ci/cd scenario, it will do its job as the `update` is a very neat feature for installing the package's dependencies in a controlled way.

NEWS.md
maksymiuks: right, done!

gaborcsardi:
Thanks!

Issue #756, opened 2023-05-10
local_package_deps: throw clear warning when `pkgdir` is not valid?


MatthieuStigler:
This is a small detail, but it would be nice to have `local_package_deps` fail in a cleaner way when `pkgdir` is not found? 

Currently, this triggers the correct message, but only as warning, then triggering a subsequent error:

> #> Warning in normalizePath(path): path[1]="wrongpath": No such file or directory
> #> Error in if (!is_dir(path)) {: missing value where TRUE/FALSE needed

``` r
library(remotes)
packageVersion("remotes")
#> [1] '2.4.2.9000'
local_package_deps("wrongpath")
#> Warning in normalizePath(path): path[1]="wrongpath": No such file or directory
#> Error in if (!is_dir(path)) {: missing value where TRUE/FALSE needed
```

<sup>Created on 2023-05-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

gaborcsardi:
A PR is welcome!

MatthieuStigler:
ok, just added `mustWork = TRUE` in `normalizePath`, now warning becomes error

PR #772, opened 2023-11-02
local_package_deps: turn warning into error 

MatthieuStigler:
local_package_deps("wrongpath") will now fail for wrong/inexisting directory, instead of throwing warnign then unrealted error, see #756


gaborcsardi:
Thanks!

Issue #782, opened 2024-02-26
Make parsing Additional_repositories optional


maksymiuks:
Hi,

I want to suggest an enhancement to how dependencies are identified in the  `dev_package_deps` function. Currently, `Additional_repositories` are always added to repos:

https://github.com/r-lib/remotes/blob/6fe8836e95805020b5f7c5b7cd2cf1321828efb7/R/deps.R#L130

which might be problematic when preparing CRAN releases. Always appending `Additional_repositories` to the repos, means that non-CRAN repos will be used for identifying  Imports/Depends dependencies when testing the package on CI/CD. Such behaviour is not allowed when undergoing CRAN checks, as only Suggests packages are allowed to pull from `Additional_repositories` then. 

Therefore I'd like to suggest adding a parameter which would make `Additional_repositories` conditional on a parameter.

I'll prepare a PR adding this parameter

maksymiuks:
PR ready https://github.com/r-lib/remotes/pull/783

maksymiuks:
Closing as the PR was merged

PR #783, opened 2024-02-26
Add additional_repositories to dev_package_deps

maksymiuks:
As showcased in the https://github.com/r-lib/remotes/issues/782 there is a PR with suggested implementation.

ping @gaborcsardi


gaborcsardi:
Thanks!

Issue #818, opened 2024-11-14
Unable to install public R packages on GitHub if SAML turned on (GitHub Enterprise)


eeholmes:
## Background

We have R packages on public repositories on GitHub Enterprise. SAML is turned on. But the repos are fully public and there are no restrictions to downloading, cloning or forking. But if we try to install the packages with `install_github()`, we get a error if the user has a PAT. Deleting the PAT will cause installation to work. The error is only with packages on GitHub Enterprise with SAML turned on.

## Reprex

1. Make sure a PAT is created. 
```
usethis::create_github_token() 
credentials::set_github_pat("YourPAT")
```
2. Try to install package
This produces the error:
```
remotes::install_github("nmfs-ost/satf")
Using GitHub PAT from the git credential store.
Error: Failed to install 'unknown package' from GitHub:
  HTTP error 403.
  Resource protected by organization SAML enforcement. You must grant your Personal Access token access to this organization.

  Rate limit remaining: 4986/5000
  Rate limit reset at: 2024-11-14 17:49:00 UTC
```
3. Delete the PAT
```
gitcreds::gitcreds_delete()
```
Now `remotes::install_github("nmfs-ost/satf")` works

4. Test on a fork that is not in Enterprise w SAML
```
remotes::install_github("eeholmes/satf")
```
works when there is a PAT.

## Where the problem appears

```
curl::curl_fetch_memory(url, handle = h)
```
called in 
```
remotes:::github_DESCRIPTION()
```
The values are
```
url= "https://api.github.com/repos/nmfs-ost/satf/contents/DESCRIPTION?ref=HEAD"
```
headers
```
Accept 
"application/vnd.github.v3.raw" 
Authorization 
"token xxxx"
```

gaborcsardi:
Did you try what the error message suggests?

eeholmes:
That is only possible for organization members. These are R package hosted for public use on GitHub Enterprise. Non-members have no way to authorize PAT for organization use. In contrast, for org members you see a button that says "Configure SSO" or something like that.

gaborcsardi:
>  no one can install our packages without deleting their PATs

That's not what I see, I do have a PAT and can install them:
```
❯ remotes::install_github("nmfs-ost/satf")
Using GitHub PAT from the git credential store.
Downloading GitHub repo nmfs-ost/satf@HEAD
These packages have more recent versions available.
It is recommended to update all of them.
Which would you like to update?
...
```

[...]

PR #819, opened 2024-11-22
respect auth_token=NULL when used in install_github()

eeholmes:
#818 SAML sign-on behavior with GitHub Enterprise makes `install_github()` impossible to use for GitHub repos on Enterprise accounts with multiple orgs. Users on one org will be unable to install R packages in public repos on other orgs in the Enterprise account.  If remotes would respect `auth_token=NULL` then that can be put in the install instructions.

Currently when you use `install_github(..., auth_token=NULL)`, remotes does not fully respect the 'request' not to use a PAT and not to use `github_pat()`. The call to `github_pat()` (instead of using NULL that was passed in) happens in 2 places.

The changes suggested cause `auth_token=NULL` to be used consistently, as if there were no PAT at all (which is what is being requested.

Since NULL is what `github_pat()` returns if there is no PAT stored, it seems like this does not break anything. It simply respects `auth_token=NULL`.


gaborcsardi:
Thanks! I am not entirely sure if that functions you changes are called from another code path or not. E.g. during `update(pkg_deps())`. If they are, this PR potentially breaks those usages if they rely on the auth token being filled in when `NULL`.

